PR #17, opened 2015-05-04
Updating to v2 of the surveys API

Erin-Mounts:
and bumping SDK version to 3


jwe-apple:
rs=me

Can't vouch for correctness but it's internally consistent.

PR #18, opened 2015-05-07
Don't crash--complete with error instead

Erin-Mounts:
...when failing to retrieve valid upload request headers for  an S3 pre-signed URL


peculiar:
You probably shouldn't include BridgeSDK.xcodeproj/xcuserdata/erin.xcuserdatad/xcdebugger/Breakpoints_v2.xcbkptlist in the commit, especially since it has absolute paths.

Erin-Mounts:
ignoring breakpoints file instead...

peculiar:
r=me

PR #20, opened 2015-06-05
Auto-refresh expired session tokens in Bridge iOS SDK

Erin-Mounts:
...and then retry the original request with the new session token.

### DIFF
--- a/BridgeSDK/BridgeAPI/SBBAuthManagerInternal.h
+++ b/BridgeSDK/BridgeAPI/SBBAuthManagerInternal.h
@@ -37,11 +37,12 @@ extern SBBEnvironment gSBBDefaultEnvironment;
 @protocol SBBAuthManagerInternalProtocol <SBBAuthManagerProtocol>
 
 - (BOOL)isAuthenticated;
+- (void)setSessionToken:(NSString *)sessionToken;
 - (void)clearSessionToken;
 
 @end
 
-@interface SBBAuthManager(internal)
+@interface SBBAuthManager(internal)<SBBAuthManagerInternalProtocol>
 
 - (void)clearKeychainStore;
 
--- a/BridgeSDK/BridgeAPI/SBBAuthManagerInternal.h
+++ b/BridgeSDK/BridgeAPI/SBBAuthManagerInternal.h
@@ -34,6 +34,13 @@
 
 extern SBBEnvironment gSBBDefaultEnvironment;
 
+@protocol SBBAuthManagerInternalProtocol <SBBAuthManagerProtocol>
+
+- (BOOL)isAuthenticated;
+- (void)clearSessionToken;
+
+@end
+
 @interface SBBAuthManager(internal)
 
 - (void)clearKeychainStore;
--- a/BridgeSDK/Networking/SBBNetworkManager.m
+++ b/BridgeSDK/Networking/SBBNetworkManager.m
@@ -376,6 +376,12 @@ NSString *kAPIPrefix = @"webservices";
 /*********************************************************************************/
 #pragma mark - Helper Methods
 /*********************************************************************************/
+
+- (NSDictionary *)headersPreparedForRetry:(NSDictionary *)headers
+{
+    return headers;
+}
+
 - (NSURLSessionDataTask *) doDataTask: (NSString*) method
                           retryObject: (APCNetworkRetryObject*) retryObject
                             URLString: (NSString*)URLString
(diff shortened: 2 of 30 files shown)


BridgeSDK/BridgeAPI/SBBAuthManagerInternal.h
```diff
@@ -34,7 +34,15 @@
...
-@interface SBBAuthManager(internal)
+@protocol SBBAuthManagerInternalProtocol <SBBAuthManagerProtocol>
+
+- (BOOL)isAuthenticated;
+- (void)setSessionToken:(NSString *)sessionToken;
+- (void)clearSessionToken;
+
+@end
+
+@interface SBBAuthManager(internal)<SBBAuthManagerInternalProtocol>
```
jwe-apple: Does this actually need to be a named category?

BridgeSDK/Networking/SBBNetworkManager.m
```diff
@@ -692,6 +683,16 @@ - (void)handleError:(NSError*)error task:(NSURLSessionDataTask*) task retryObjec
     }
 }
 
+- (void)handleHTTPError:(NSError *)error task:(NSURLSessionDataTask *)task retryObject:(APCNetworkRetryObject *)retryObject
+{
+    //TODO: Add retry for Server maintenance
```
jwe-apple: Do you have a bug tracking this?

BridgeSDKSample/BridgeSDKSample/AppDelegate.m
```diff
@@ -69,4 +79,35 @@ - (void)application:(UIApplication *)application handleEventsForBackgroundURLSes
...
+{
+    [[NSUserDefaults standardUserDefaults] setObject:sessionToken forKey:kSessionTokenKey];
+}
+
+- (void)authManager:(id<SBBAuthManagerProtocol>)authManager didGetSessionToken:(NSString *)sessionToken forUsername:(NSString *)username andPassword:(NSString *)password
+{
+    [self authManager:authManager didGetSessionToken:sessionToken];
+    [[NSUserDefaults standardUserDefaults] setObject:username forKey:kUsernameKey];
+    
+    // ***** REALLY BAD IDEA. DEMO ONLY. DON'T EVER DO THIS IN A REAL APP. *****
```
jwe-apple: Same comment also applies to the session token!

jwe-apple:
r=me with extremely minor comments.

jwe-apple:
@Erin-Mounts Presumably we need to be ensuring that latest version of apps applies this. Are the apps still running off a branch?

peculiar:
@jwe-apple @Erin-Mounts The apps are still using the last commit on the `rk-compat` branch. This branch has some modifications to the survey feature that the apps require. Any reason why the compat branch could not be merged into master?

Erin-Mounts:
The apps shouldn’t require any of those mods anymore. The SBBSurvey “questions” property has been replaced in the Bridge API with the “elements” property, and the SBBSurveyQuestion “detail” property was given the name “promptDetail” when it was added to the API for internal consistency. We didn’t want to upset the apple cart before the March 9 launch, so we didn’t address it at the time, but these should be pretty straightforward to fix in the apps now—search and replace, mostly. I’d rather not incorporate what were meant to be temporary hacks that don’t reflect the underlying API into the main branch and have to support them going forward.

From my point of view the only change in the rk-compat branch that belongs in master is changing the “only active architectures” build flag from NO to YES to fix a warning.

BridgeSDK/BridgeAPI/SBBAuthManagerInternal.h
Erin-Mounts: The unnamed "private" category is already being used in the .m file, and I didn't want to move everything from there to here. So, yes.

BridgeSDK/Networking/SBBNetworkManager.m
Erin-Mounts: Actually no, I believe this is a comment that was in the original networking code provided to us by YML and now that you bring it up, I'm not sure it's relevant anymore--this won't get called for background NSURLSession tasks, and likely isn't really appropriate when the UI might be waiting on a response.

BridgeSDKSample/BridgeSDKSample/AppDelegate.m
Erin-Mounts: I suppose so. I'll add it.

PR #21, opened 2015-06-10
Fix background upload bug introduced with previous PR

Erin-Mounts:
Make sure SBBBridgeNetworkManager (not SBBNetworkManager) default instance is the delegate of the global background session


peculiar:
r=me

PR #22, opened 2015-06-16
log upload IDs to make diagnosing upload issues easier

DwayneJengSage:
Logging upload IDs (on either success or failure) would be very useful to debug data issues (not just for YML, but for us as well). I made a code change to add the logging statements. Please review and/or let me know if there's a different approach that would work better.

Testing done: Launched BridgeSDKSample, uploaded a file, verified expected log entries in the log.

### DIFF
--- a/BridgeSDK/BridgeAPI/SBBUploadManager.m
+++ b/BridgeSDK/BridgeAPI/SBBUploadManager.m
@@ -370,6 +370,7 @@ static NSString *kUploadSessionsKey = @"SBBUploadSessionsKey";
   }
   [self setUploadRequestJSON:nil forFile:downloadTask.taskDescription];
   if ([uploadSession isKindOfClass:[SBBUploadSession class]]) {
+    NSLog(@"Successfully obtained upload session with upload ID %@", uploadSession.id);
     [self setUploadSessionJSON:jsonObject forFile:downloadTask.taskDescription];
     NSDictionary *uploadHeaders =
     @{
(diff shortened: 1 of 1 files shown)


Erin-Mounts:
Just two things:
1) all the code that logs to the console should be bracketed by #if DEBUG on the line before, and #endif on the line after, so it doesn't log to the console in production builds.
2) When the request results in an error, you should include the error object in the debug log output. E.g.,  NSLog(@"Error uploading to S3 for upload ID %@:\n%@", uploadSession.id, error);

Erin-Mounts:
looks good. r=me but I think we're supposed to have a pair of non-Sage eyes look at it as well before merging it in. Can you have Chris take a quick look?

BridgeSDK/BridgeAPI/SBBUploadManager.m
```diff
@@ -377,8 +380,19 @@ - (void)URLSession:(NSURLSession *)session downloadTask:(NSURLSessionDownloadTas
       @"Content-Type": uploadRequest.contentType,
       @"Content-MD5": uploadRequest.contentMd5
       };
+    SBBNetworkManagerTaskCompletionBlock uploadFileCompletion = nil;
```
peculiar: For the non-DEBUG version of this framework, how is *upload file completion* communicated if this block is nil?

peculiar:
For the purpose of logging, your changes look good.

BridgeSDK/BridgeAPI/SBBUploadManager.m
Erin-Mounts: It's a multi-step process to make background upload and confirmation appear as one seamless asynch call with a completion handler. It's actually a series of steps, each triggered by completion of the last. The completion block passed into the uploadFileToBridge:contentType:completion: method is saved, associated with the task ID. The SBBUploadManager is set as the background session delegate for SBBBridgeNetworkManager so the NSURLSession delegate methods are forwarded here. When all the steps are completed (or fail at some point along the way), the original completion block will be looked up and called with the error, if any.

PR #24, opened 2015-06-23
Add optional delegate method to SBBUploadManager for data verification

chrisinsfo:
- Optional delegate method available only in debug mode.
- Called when a background file upload task finishes, successfully or otherwise in debug mode.
- Provides a URL that can be called for more information about the post-upload status of the data.
- Intended to work with APCDataVerificationDaemon in AppCore (forthcoming) which queues URL's and calls after a time interval for verification.

### DIFF
--- a/BridgeSDK/BridgeAPI/SBBUploadManager.h
+++ b/BridgeSDK/BridgeAPI/SBBUploadManager.h
@@ -53,8 +53,8 @@ typedef void (^SBBUploadManagerCompletionBlock)(NSError *error);
 
 @optional
 /*!
- Optional delegate method available only in debug mode.
- Called when a background file upload task finishes, successfully or otherwise in debug mode.
+ Optional delegate method available only in debug builds.
+ Called when a background file upload task finishes, successfully or otherwise in debug builds.
  Provides a URL that can be called for more information about the post-upload status of the data.
  
  @param manager The upload manager instance making the call.
--- a/BridgeSDK/BridgeAPI/SBBUploadManager.h
+++ b/BridgeSDK/BridgeAPI/SBBUploadManager.h
@@ -51,6 +51,18 @@ typedef void (^SBBUploadManagerCompletionBlock)(NSError *error);
  */
 - (void)uploadManager:(SBBUploadManager *)manager uploadOfFile:(NSString *)file completedWithError:(NSError *)error;
 
+@optional
+/*!
+ Optional delegate method available only in debug mode.
+ Called when a background file upload task finishes, successfully or otherwise in debug mode.
+ Provides a URL that can be called for more information about the post-upload status of the data.
+ 
+ @param manager The upload manager instance making the call.
+ @param file    The path to the file whose upload task finished or failed.
+ @param url     The url to call to get more information about the post-upload status of the data.
+ */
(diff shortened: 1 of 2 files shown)


BridgeSDK/BridgeAPI/SBBUploadManager.h
```diff
@@ -51,6 +51,18 @@ typedef void (^SBBUploadManagerCompletionBlock)(NSError *error);
...
+ Optional delegate method available only in debug mode.
+ Called when a background file upload task finishes, successfully or otherwise in debug mode.
+ Provides a URL that can be called for more information about the post-upload status of the data.
+ 
+ @param manager The upload manager instance making the call.
+ @param file    The path to the file whose upload task finished or failed.
+ @param url     The url to call to get more information about the post-upload status of the data.
+ */
+- (void)uploadManager:(SBBUploadManager *)manager uploadOfFile:(NSString *)file completedWithVerificationURL:(NSURL *)url;
+
```
Erin-Mounts: Instead of "debug mode" (in both places it's used) I'd say "Debug builds" since it's a compile-time thing, not a runtime mode setting.

I had also been thinking this declaration and the accompanying headerdoc comments should be wrapped in the conditional compilation macros but given that it's optional, and that there's no way to enforce the implementation of the method also being wrapped in the conditional macros, I guess there's no point. So never mind on that.

chrisinsfo: done

BridgeSDK/BridgeAPI/SBBUploadManager.m
```diff
@@ -430,18 +430,24 @@ - (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didComp
...
+            NSLog(@"Error calling upload complete for upload ID %@:\n%@", uploadSession.id, error);
+        } else {
+            NSString* uploadStatusUrlString = nil;
+            if ([self.networkManager isKindOfClass:[SBBNetworkManager class]]) {
+                NSString* relativeUrl = [NSString stringWithFormat:@"/api/v1/upload/%@/status", uploadSession.id];
+                NSURL* url = [(SBBNetworkManager*) self.networkManager URLForRelativeorAbsoluteURLString:relativeUrl];
+                uploadStatusUrlString = [url absoluteString];
+                
+                if ([_uploadDelegate respondsToSelector:@selector(uploadManager:uploadOfFile:completedWithVerificationURL:)]) {
+
```
Erin-Mounts: The extraneous blank lines don't follow the whitespace conventions of the rest of the file.

Also, maybe it'd be good to fix the tab-spacing of the whole file to be consistent with this; I originally developed this on my personal laptop which I'd set for a previous job to two-space tabs, but have since standardized on four for new files, as you've used here.

Erin-Mounts: (FYI if your Xcode editor is set to 4-space tabs, just select-all/cut/paste and it will fix all the tabs in the file)

chrisinsfo: I did a ctrl + i on it. Hope that's what you wanted!

Erin-Mounts: Looks right. r=me

Erin-Mounts:
r=me

PR #25, opened 2015-07-24
Integration test automation

Erin-Mounts:
Includes initial round of automated integration tests, and fixes for a few bugs uncovered as a result; this will constitute the final iteration of v2 before moving on to v3 API support (and will be branched as v2final).


BridgeSDK/Networking/SBBNetworkManager.m
```diff
@@ -218,6 +218,11 @@ - (instancetype) initWithBaseURL: (NSString*) baseURL bridgeStudy: (NSString*)br
         self.environment = SBBEnvironmentCustom;
```
EricSiegNW: Is this right? Does it get reset somewhere else? I realize its not part of the PR changes but something I noticed

BridgeSDK/Networking/SBBNetworkManager.m
```diff
@@ -218,6 +218,11 @@ - (instancetype) initWithBaseURL: (NSString*) baseURL bridgeStudy: (NSString*)br
...
         self.uploadCompletionHandlers = [NSMutableDictionary dictionary];
         self.downloadCompletionHandlers = [NSMutableDictionary dictionary];
+        if (bridgeStudy.length) {
```
EricSiegNW: Could you add a couple word comment explaining what is going on here?

BridgeSDKTests/SBBBridgeAPIUnitTestCase.m
```diff
@@ -1,23 +1,31 @@
...
 
-#import "SBBBridgeAPITestCase.h"
+#import "SBBBridgeAPIUnitTestCase.h"
 #import "SBBAuthManagerInternal.h"
 #import "SBBBridgeNetworkManager.h"
 #import "SBBNetworkManagerInternal.h"
+#import "TestAdminAuthDelegate.h"
 
-@implementation SBBBridgeAPITestCase
+@interface SBBBridgeAPIUnitTestCase ()
```
EricSiegNW: NIce differentiation, also excited to see integration tests :)

EricSiegNW:
The code looks good. However, when I try to run the BridgeSDKIntegrationTests suite I get a build failure: The file "BridgeAdminCredentials.plist" couldn't be opened because there is no such file. Does that need to be checked in? I'm guessing it might be something that will exist in the CI environment rather than being checked in, but still wanted to call it out.

BridgeSDK/Networking/SBBNetworkManager.m
Erin-Mounts: Yes, this is only a default setting. This initializer isn't exposed in the public API and callers reset it appropriately.

BridgeSDK/Networking/SBBNetworkManager.m
Erin-Mounts: done

Erin-Mounts:
Yes, as you might guess, that .plist contains admin account credentials for the integration tests to use, and when Alx sent them to me he suggested they were super sensitive and I should eat my computer after seeing them... so no, they won't be checked into a public Github repo any time soon... ;)

FYI the convention between Apple, Y Media, and Sage when reviewing PRs to any of the repos involved in ResearchKit (of which this is, peripherally, one) is for the reviewer to comment "r=me" when they're satisfied, before the PR is accepted.

EricSiegNW:
r=me

PR #26, opened 2015-07-27
V3 updates

Erin-Mounts:
Updating to use v3 Bridge REST APIs. Note especially that SBBProfileManager has been renamed to SBBUserManager to reflect the API change, and that the data sharing scope settings have been moved from SBBConsentManager to SBBUserManager (again, to reflect the API changes).


BridgeSDK/BridgeAPI/SBBAuthManager.m
```diff
@@ -267,17 +277,17 @@ - (NSString *)sessionToken
 
 - (NSURLSessionDataTask *)signUpWithEmail:(NSString *)email username:(NSString *)username password:(NSString *)password completion:(SBBNetworkManagerCompletionBlock)completion
 {
-    return [_networkManager post:@"/api/v1/auth/signUp" headers:nil parameters:@{@"study":gSBBAppStudy, @"email":email, @"username":username, @"password":password, @"type":@"SignUp"} completion:completion];
+    return [_networkManager post:kSBBAuthSignUpAPI headers:nil parameters:@{@"study":gSBBAppStudy, @"email":email, @"username":username, @"password":password, @"type":@"SignUp"} completion:completion];
```
EricSiegNW: Love the change to make the uri a constant. Is there any way to do something similar for the parameters? Some sort of mapping approach? It seems like a parameter name change in the future will be more painful and vulnerable to errors down the line. For example RestKit offers a convention based mapping to convert api parameters to object variables that I had liked in the past. I don't have the context to know if that has any relevance here and the typed out approach is something I see a lot, just wanted to highlight this. Definitely not a blocker.

EricSiegNW:
r=me

BridgeSDK/BridgeAPI/SBBAuthManager.m
Erin-Mounts: It's a fair point. I've got a convention where I create a model object with the same name as the API "type" field, prepended with SBB, and then give the fields of the object the same names as the JSON fields. I could have created an SBBSignUp object, filled it in (getting compile-time field name and type checking), and then asked the SBBObjectManager for the Bridge JSON for that object. But that only moves the fragility problem to the object model definitions. Also since we never receive SignUp objects from Bridge, only send them, we likely wouldn't notice server-side field name changes anyway until this call started failing, so we kind of have to rely on integration testing to catch stuff like this. And since it didn't seem like it would help with that, it seemed like overkill to create an SDK object and convert it to JSON at runtime when it's only ever used in this one call.

PR #27, opened 2015-07-28
SBBProfileManager -> SBBUserManager rename

EricSiegNW:
I think this is a rename that slipped through earlier? Not sure why it didn't show up as a problem until now.


Erin-Mounts:
Whoops, yeah. Tests and sample app probably include that header directly so didn't notice it missing from the module header...

r=me

PR #32, opened 2015-09-28
Update Task API parameters

Erin-Mounts:
and add Task Manager integration test
and update Task Manager unit test
and fix integration tests to run on device as well as simulator


EricSiegNW:
Approved!

r=me

PR #33, opened 2015-10-01
Make NSDate category public

Erin-Mounts:


EricSiegNW:
I'm not sure all the xcscheme changes make sense? When I made the same change, my only modified files were the first 2 small changes you have.

Feel free to merge tho, as we work on testing/figuring this out.

r=me

Erin-Mounts:
yeah not sure where those came from, unless as a side effect of running integration tests…

PR #36, opened 2015-10-07
Fix crasher in SBBNetworkManager.m

### LINKED ISSUE #34: nil data object in SBBNetworkManager causes a crash.
I'm working on the MyHeartCounts app and I noticed a crasher while I was flying from NY back to SJ.
On line 420 of the SBBNetworkManager, if the data object that is being used in

NSDictionary * responseObject = [NSJSONSerialization JSONObjectWithData:data options:0 error:NULL];

is nil the application will crash.

Erin-Mounts:
Fixes #34

### DIFF
--- a/BridgeSDK/Networking/SBBNetworkManager.m
+++ b/BridgeSDK/Networking/SBBNetworkManager.m
@@ -417,7 +417,15 @@ NSString *kAPIPrefix = @"webservices";
   NSMutableURLRequest *request = [self requestWithMethod:method URLString:URLString headers:headers parameters:parameters error:nil];
     __block NSURLSessionDataTask *task = [self.mainSession dataTaskWithRequest:request completionHandler:^(NSData *data, NSURLResponse *response, NSError *error) {
         NSError * httpError = [NSError generateSBBErrorForStatusCode:((NSHTTPURLResponse*)response).statusCode data:data];
-        NSDictionary * responseObject = [NSJSONSerialization JSONObjectWithData:data options:0 error:NULL];
+        id responseObject = nil;
+        if (data.length) {
+            responseObject = [NSJSONSerialization JSONObjectWithData:data options:0 error:NULL];
+            if (!responseObject) {
+                // maybe it was an html error page
+                responseObject = [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
+            }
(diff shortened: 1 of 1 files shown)


BridgeSDK/Networking/SBBNetworkManager.m
```diff
@@ -417,7 +417,15 @@ - (NSURLSessionDataTask *) doDataTask: (NSString*) method
   NSMutableURLRequest *request = [self requestWithMethod:method URLString:URLString headers:headers parameters:parameters error:nil];
     __block NSURLSessionDataTask *task = [self.mainSession dataTaskWithRequest:request completionHandler:^(NSData *data, NSURLResponse *response, NSError *error) {
         NSError * httpError = [NSError generateSBBErrorForStatusCode:((NSHTTPURLResponse*)response).statusCode data:data];
-        NSDictionary * responseObject = [NSJSONSerialization JSONObjectWithData:data options:0 error:NULL];
+        id responseObject = nil;
+        if (data.length) {
+            responseObject = [NSJSONSerialization JSONObjectWithData:data options:0 error:NULL];
+            if (!responseObject) {
+                // maybe it was an html error page
+                responseObject = [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
```
EricSiegNW: If it was an html error page, should it check for an httpError and create one if none exists? What if the server returns something else?

Current approach seems like there could be a problem without either of the error conditionals below getting called since this acts to handle an error like situation without actually creating an NSError object or validating that an NSError has been created that will later be handled.

Erin-Mounts: So, there's the situation where the iOS network call returns an error, which means there was a network error; and then if the request goes through and gets any kind of actual response  (including an http error response) it's up to the calling code to detect and handle that. I do this outside of the code block that checks for and handles errors so that the decoded response can be passed through to both "good" and "bad" paths. There's nothing here that will prevent proper error handling down the road, it just allows for that error handling code to log whatever might have come back, which in some odd cases I've seen to be raw html rather than json even when it hits one of our servers.

EricSiegNW: Makes sense.

r=me

PR #38, opened 2015-10-09
Update Bridge sample app to work with changed Task Manager API

Erin-Mounts:
also a tweak to handle an occasional situation on the signIn/signUp page where the keyboard blocks the signIn/signUp button and can't be dismissed; also also inadvertently included a commit to fix how we handle getting the app name and version for the UserAgent http header, due to Apple apparently changing the behavior of the localizedInfoDictionary NSBundle method as of iOS 8.3, so switching to recommended API for accessing the required info that I should have been using in the first place


EricSiegNW:
r=me

PR #42, opened 2015-11-11
Fix consent and activities to pass int. tests

Erin-Mounts:
also test other time zone offsets besides PST for scheduled activities API


EricSiegNW:
r=me

PR #43, opened 2015-11-11
Fix activity manager integration tests

Erin-Mounts:
just since last night the researcher API changed in a way that broke the existing tests. also made it robust against multiple tests being run simultaneously.


EricSiegNW:
r=me

PR #44, opened 2015-11-20
BRIDGE-652 Kill switch for old versions

syoung-smallwisdom:
Mapped the 410 (Gone) http status code returned by the server for app versions that are no longer supported to a BridgeSDK error code.  

Added NSLocalizedString macro to facilitate find-and-replace at some point in the future when error messages might need to be localized.

Deprecated SBBErrorCodes and replaced with SBBErrorCode which uses the enum format required for compatibility with Swift 2.0 enums. The use of the "k" prefix to mark a constant has gone out of vogue.

Handling the new 410 error code will be done in the mPower app (or possibly the AppCore). Will be implementing that in a future PR in that repository.

### DIFF
--- a/BridgeSDK/BridgeAPI/Categories/NSError+SBBAdditions.m
+++ b/BridgeSDK/BridgeAPI/Categories/NSError+SBBAdditions.m
@@ -160,7 +160,8 @@
 
 + (NSError *)generateSBBObjectNotExpectedClassErrorForObject:(id)object expectedClass:(Class)expectedClass
 {
-  NSString *localizedFormat = NSLocalizedString(@"Object '%1$@' is of class %2$@, expected class %3$@", @"Error Description");
+  NSString *localizedFormat = NSLocalizedString(@"Object '%1$@' is of class %2$@, expected class %3$@",
+                                                @"Error Description: Object is not of expected class. object description=$1, actual class=$2, expected class=$3");
   NSString *desc = [NSString stringWithFormat:localizedFormat, object, NSStringFromClass([object class]), NSStringFromClass(expectedClass)];
   return [NSError errorWithDomain:SBB_ERROR_DOMAIN code:SBBErrorCodeObjectNotExpectedClass userInfo:@{NSLocalizedDescriptionKey: desc}];
 }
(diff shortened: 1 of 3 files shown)


BridgeSDK/BridgeAPI/Categories/NSError+SBBAdditions.m
```diff
@@ -75,54 +84,85 @@ + (NSError*) generateSBBErrorForStatusCode:(NSInteger)statusCode data: (id) data
...
+  NSString *localizedFormat = NSLocalizedString(@"Error reading temp file for original file URL:\n%@", @"Error Description: error reading temp file");
+  NSString *desc = [NSString stringWithFormat:localizedFormat, url];
+  return [NSError errorWithDomain:SBB_ERROR_DOMAIN code:SBBErrorCodeTempFileError userInfo:@{NSLocalizedDescriptionKey: desc}];
 }
 
 + (NSError *)generateSBBObjectNotExpectedClassErrorForObject:(id)object expectedClass:(Class)expectedClass
 {
-  NSString *desc = [NSString stringWithFormat:@"Object '%@' is of class %@, expected class %@", object, NSStringFromClass([object class]), NSStringFromClass(expectedClass)];
-  return [NSError errorWithDomain:SBB_ERROR_DOMAIN code:kSBBObjectNotExpectedClass userInfo:@{NSLocalizedDescriptionKey: desc}];
+  NSString *localizedFormat = NSLocalizedString(@"Object '%1$@' is of class %2$@, expected class %3$@", @"Error Description");
```
Erin-Mounts: When localizing format strings, I like to explain in the description what parameters it expects, e.g. "Error Description, to be filled in with an object pointer, the classname of that object, and the classname that was expected" so the translator has a better idea how to phrase the translation.  (Also by default the genstrings tool 'positionalizes' the parameters for you when generating the strings file so it's optional to do it explicitly here, but fine.)

syoung-smallwisdom: good point. it was kind of a weird error to try and describe and i wasn't sure what to say. thoughts?

Erin-Mounts: Well, this particular error is one that shouldn't happen in production and should never be shown to an end user (maybe used in error logging though). So maybe something like "Message for internal error logging when an object of unexpected class is received from the server, to be filled in with..." etc. as in my previous comment.

Erin-Mounts:
Awesome. r=me

PR #46, opened 2015-11-25
Localize BridgeSDK error messages

Erin-Mounts:
...into our own .strings file with its own name so someday when we support BridgeSDK being a CocoaPod it won't conflict with the app's Localizable.strings file when its gets pulled into the main bundle, which apparently is a thing that happens, according to a commit comment I saw on ResearchKit...


EricSiegNW:
r=me

PR #49, opened 2015-12-14
Removed category override in AppCore for returning custom SBBSchedule description

syoung-smallwisdom:
There was an category override in AppCore of SBBSchedule to implement a description. I have moved this to BridgeSDK.

### DIFF
--- a/BridgeSDK/BridgeAPI/BridgeAPITypes/SBBBridgeObject.m
+++ b/BridgeSDK/BridgeAPI/BridgeAPITypes/SBBBridgeObject.m
@@ -49,4 +49,9 @@
   return self;
 }
 
+- (NSString *)description
+{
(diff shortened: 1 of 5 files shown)


BridgeSDK/BridgeAPI/BridgeAPITypes/SBBSchedule.m
```diff
@@ -33,6 +33,19 @@ @implementation SBBSchedule
 
 #pragma mark Abstract method overrides
 
-// Custom logic goes here.
+- (NSString *) description
```
Erin-Mounts: It makes sense to provide "description" methods for our SBBxxx classes and not make AppCore implement them. Rather than doing it in each class though, I'd suggest putting a description method on the parent SBBBridgeObject class, that (for example) returns [[self dictionaryRepresentation] description] and get all of them covered at once.

PR #50, opened 2016-01-12
Add support for new multiple consents and subpopulations APIs

Erin-Mounts:
and integration testing (moved to its own test study to avoid collisions with server-side integration tests)

### DIFF
--- a/BridgeSDK/BridgeAPI/SBBBridgeNetworkManager.m
+++ b/BridgeSDK/BridgeAPI/SBBBridgeNetworkManager.m
@@ -123,6 +123,7 @@
         }];
     } else {
         [self checkForAndHandleUnsupportedAppVersionHTTPError:error];
+        [self checkForAndHandleServerPreconditionNotMetHTTPError:error task:task responseObject:responseObject retryObject:retryObject];
         [super handleHTTPError:error task:task response:responseObject retryObject:retryObject];
     }
 }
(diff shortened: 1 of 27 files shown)


BridgeSDK/BridgeAPI/SBBBridgeNetworkManager.m
```diff
@@ -163,6 +164,26 @@ - (void)checkForAndHandleUnsupportedAppVersionHTTPError:(NSError *)error
...
+        // implementing this method and not coincidentally using the same method signature for something else.
+        id appDelegate = [[UIApplication sharedApplication] delegate];
+        if (![appDelegate conformsToProtocol:@protocol(SBBBridgeAppDelegate)] ||
+            ![appDelegate respondsToSelector:@selector(handleUserNotConsentedError:sessionInfo:networkManager:)] ||
+            ![appDelegate handleUserNotConsentedError:error sessionInfo:responseObject networkManager:self])
+        {
+#if DEBUG
+            // Log the error to the console
+            NSLog(@"User Not Consented error not handled by app delegate:\n%@", error);
+#endif
```
syoung-smallwisdom: app version handler has a fall-through for the case where the app delegate does not handle it. What happens in this case?

syoung-smallwisdom:
One question. Otherwise, looks good.

BridgeSDK/BridgeAPI/SBBBridgeNetworkManager.m
Erin-Mounts: There didn’t seem like a reasonable default way to handle it in the SDK as there is with 410-you-need-to-upgrade, since it requires study-specific UI to get the user to re-consent, so it just logs the error to the console and (in any case) ends up sending the error to the completion handler, which is where it’s currently handled by existing apps anyway.

EricSiegNW:
Looks good, also seems to be working great with where we've used it already.

PR #52, opened 2016-02-25
Prevent crashes when accessing SBBUploadManager from multiple threads

pmkowal:
Hi, I'm currently using `Bridge-iOS-SDK` in my project and I found that the application that I'm working on crashes when `SBBUploadManager` instance is accessed from multiple threads at the same time. 

Please consider the following solution as it should prevent from crashes and make the `SBBUploadManager` class thread-safe.

**Changelog**:

- add `@synchronized` block when accessing `uploadCompletionHandlers` dictionary

### DIFF
--- a/BridgeSDK/BridgeAPI/SBBUploadManager.m
+++ b/BridgeSDK/BridgeAPI/SBBUploadManager.m
@@ -181,10 +181,8 @@ static NSString *kUploadSessionsKey = @"SBBUploadSessionsKey";
 
 - (SBBUploadManagerCompletionBlock)completionBlockForFile:(NSString *)file
 {
-    @synchronized(self) {
-        SBBUploadCompletionWrapper *wrapper = [_uploadCompletionHandlers objectForKey:file];
-        return wrapper.completion;
-    }
+    SBBUploadCompletionWrapper *wrapper = [_uploadCompletionHandlers objectForKey:file];
+    return wrapper.completion;
 }
 
 - (void)setCompletionBlock:(SBBUploadManagerCompletionBlock)completion forFile:(NSString *)file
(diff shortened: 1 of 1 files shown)


Erin-Mounts:
Good catch, thanks! As far as I can tell you've got a valid solution here. However, @synchronized is relatively heavyweight, blocks until available, and is prone to deadlocking if you're not careful with your call trees (though admittedly that's not going to be an issue here) so I generally try to avoid it in favor of serializing access via serial queues (NSOperationQueues or GCD queues) when possible. And it turns out there's a pretty easy way to do that here.

Except in two places, the uploadCompletionHandlers dictionary is only ever accessed directly from NSURLSession callbacks and delegate handlers, and so will already always be executed on the background session's (serial) delegate queue. So instead of using @synchronized as you've done, I'd suggest just wrapping those two exceptions in this:
`    [((SBBNetworkManager *)self.networkManager).backgroundSession.delegateQueue addOperationWithBlock:^{
        ...
    }];
`
so they'll be executed on that same serial queue.

The first place that needs to be tossed onto that queue is this line (currently line 280 of SBBUploadManager.m) in  `uploadFileToBridge:contentType:completion:`:
`[self setCompletionBlock:completion forFile:[tempFileURL path]];`

The other place is this line (line 459) inside the network manager call completion block in `URLSession:task:didCompleteWithError:`:
`[self completeUploadOfFile:uploadTask.taskDescription withError:error];`

If you prefer, I can just make those changes and push them to the repo.

pmkowal:
:+1: Thanks for detailed explanation!
I see your point.
> If you prefer, I can just make those changes and push them to the repo.

Yes, sure.

Erin-Mounts:
Thanks!
r=me

PR #54, opened 2016-03-08
Only release the reachability object if for some reason alloc init returns nil

syoung-smallwisdom:
Oops.


Erin-Mounts:
Looks good

PR #56, opened 2016-03-23
fix for : [__NSCFDictionary removeObjectForKey:]: attempt to remove nil key

dariuszlesniak:
i am trying to prevent some of the crashes in our production app caused by nil fileURLString, i was hoping you might help identifying cases in which fileURLString might be nil. 
Here is the stacktrace for the crash:

Thread : Fatal Exception: NSInvalidArgumentException
0  CoreFoundation                 0x182022e38 __exceptionPreprocess
1  libobjc.A.dylib                0x181687f80 objc_exception_throw
2  CoreFoundation                 0x182022d80 -[NSException initWithCoder:]
3  CoreFoundation                 0x181f46840 -[__NSCFDictionary removeObjectForKey:]
4  BridgeSDK                      0x10132c588 -[SBBUploadManager setUploadRequestJSON:forFile:] (SBBUploadManager.m:206)
5  BridgeSDK                      0x10132d1a8 -[SBBUploadManager completeUploadOfFile:withError:] (SBBUploadManager.m:320)
6  BridgeSDK                      0x10132d810 -[SBBUploadManager URLSession:downloadTask:didFinishDownloadingToURL:] (SBBUploadManager.m:371)
7  BridgeSDK                      0x10133896c -[SBBNetworkManager URLSession:downloadTask:didFinishDownloadingToURL:] (SBBNetworkManager.m:588)
8  CFNetwork                      0x182731bac __82-[NSURLSession delegate_downloadTask:didFinishDownloadingToURL:completionHandler:]_block_invoke
9  Foundation                     0x1829dc510 __NSBLOCKOPERATION_IS_CALLING_OUT_TO_A_BLOCK__
10 Foundation                     0x18292e900 -[NSBlockOperation main]
11 Foundation                     0x18291eed8 -[__NSOperationInternal _start:]
12 Foundation                     0x1829de904 __NSOQSchedule_f
13 libdispatch.dylib              0x181a6d47c _dispatch_client_callout
14 libdispatch.dylib              0x181a794c0 _dispatch_queue_drain
15 libdispatch.dylib              0x181a70f80 _dispatch_queue_invoke
16 libdispatch.dylib              0x181a7b390 _dispatch_root_queue_drain

### DIFF
--- a/BridgeSDK/BridgeAPI/SBBUploadManager.m
+++ b/BridgeSDK/BridgeAPI/SBBUploadManager.m
@@ -202,6 +202,11 @@ static NSString *kUploadSessionsKey = @"SBBUploadSessionsKey";
 
 - (void)setUploadRequestJSON:(id)json forFile:(NSString *)fileURLString
 {
+    if (!fileURLString) {
+        NSLog(@"Aborting setting json on nil fileURLString");
+        return;
+    }
+    
     NSUserDefaults *defaults = [NSUserDefaults standardUserDefaults];
     
     NSMutableDictionary *uploadRequests = [[defaults dictionaryForKey:kUploadRequestsKey] mutableCopy];


Erin-Mounts:
It gets the fileURLString from the taskDescription on the download task object. The only way it could be nil is if the task's taskDescription property wasn't set, and I can see only two ways for that to happen, one of which is that you're calling SBBNetworkManager's `downloadFileFromURLString:method:httpHeaders:parameters:taskDescription:downloadCompletion:taskCompletion:` method directly with nil for the taskDescription. 

Assuming that's not the case, then the other possibility (and I'm wondering why I haven't seen this crash before--is this new behavior in iOS 9.3?) is that iOS is cleaning up the temp directory in between when the file to be uploaded is copied into it and when (later in the same method) it takes the path of the URL to the temp file to use as the taskDescription.

If you can confirm that you're in fact *not* calling the aforementioned SBBNetworkManager method directly with a nil taskDescription, I'll just fix it to use the documents directory instead of the system temp dir for temporarily storing the file to be uploaded.

Erin-Mounts:
Closing in favor of https://github.com/Sage-Bionetworks/Bridge-iOS-SDK/pull/57.

dariuszlesniak:
do you think the system would clean the file from tmp directory during execution of a method that created it ? , that would be quite strange behaviour in normal situation, from what i found tmp directory should be cleaned while the app is not running

Erin-Mounts:
Agreed, it seems pretty unlikely, in light of what little documentation I've been able to find about the precise behavior of the app-specific tmp directory. But unless the same background session is somehow being used for non-Bridge-upload purposes and not setting a taskDescription on its tasks, I'm not seeing any other way for that fileURLString to end up as nil.

PR #57, opened 2016-03-23
Don't use system temp dir for file uploads

Erin-Mounts:
because they get deleted after three days. Instead use NSApplicationSupportDirectory.

Also, update integration tests so they work.

### DIFF
--- a/BridgeSDK/BridgeAPI/SBBUploadManager.m
+++ b/BridgeSDK/BridgeAPI/SBBUploadManager.m
@@ -113,8 +113,10 @@ static NSString *kUploadSessionsKey = @"SBBUploadSessionsKey";
 
 - (NSURL *)tempUploadDirURL
 {
-    NSURL *uploadDir = nil;
+    NSURL *appSupportDir = [[[NSFileManager defaultManager] URLsForDirectory:NSApplicationSupportDirectory inDomains:NSUserDomainMask] firstObject];
+    NSString *bundleName = [[NSBundle mainBundle].infoDictionary objectForKey:@"CFBundleIdentifier"];
     NSString *tempUploadDirName = [NSTemporaryDirectory() stringByAppendingPathComponent:@"SBBUploadManager"];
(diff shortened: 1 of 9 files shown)


syoung-smallwisdom:
Looks good

BridgeSDK/BridgeAPI/SBBUploadManager.m
```diff
@@ -113,8 +113,10 @@ - (instancetype)init
...
 - (NSURL *)tempUploadDirURL
 {
-    NSURL *uploadDir = nil;
+    NSURL *appSupportDir = [[[NSFileManager defaultManager] URLsForDirectory:NSApplicationSupportDirectory inDomains:NSUserDomainMask] firstObject];
+    NSString *bundleName = [[NSBundle mainBundle].infoDictionary objectForKey:@"CFBundleIdentifier"];
     NSString *tempUploadDirName = [NSTemporaryDirectory() stringByAppendingPathComponent:@"SBBUploadManager"];
+    NSURL *uploadDir = [[appSupportDir URLByAppendingPathComponent:bundleName] URLByAppendingPathComponent:tempUploadDirName];
     NSError *error;
     if ([[NSFileManager defaultManager] createDirectoryAtPath:tempUploadDirName withIntermediateDirectories:YES attributes:nil error:&error]) {
         uploadDir = [NSURL fileURLWithPath:tempUploadDirName];
```
dariuszlesniak: i think this is wrong, it will create the same directory as previously tmp/SBBUploadManager and that's the one that is going to be used, Library/Application Support is not used at all

Erin-Mounts: holy cow, how did I miss this? I'll submit a new PR with a fix shortly. Thanks for catching!!

Erin-Mounts: https://github.com/Sage-Bionetworks/Bridge-iOS-SDK/pull/60

PR #59, opened 2016-03-30
updating SBBUploadManager:setUploadRequestJSON:forFile to use delegateQueue

dariuszlesniak:
updating SBBUploadManager:setUploadRequestJSON:forFile to use backgroundSession.delegateQueue as concurrent executions cause dirty reads which further cause:  "Error retrieving upload request headers for temp file URL ", and eventually uploads fail


Erin-Mounts:
This method is already always called from within the background session delegate queue, with the exception of where it's called from near the end of `uploadFileToBridge:contentType:completion:` (line 307). I'd prefer you just dispatch that one call to the background session delegate queue and update this PR.

dariuszlesniak:
got it, will update PR

dariuszlesniak:
updated, please review

Erin-Mounts:
Looks good.

PR #61, opened 2016-04-13
Expose formatters used to convert date to string for json parsing

syoung-smallwisdom:


Erin-Mounts:
ask and ye shall receive ;)

PR #62, opened 2016-04-25
Refactor for caching

Erin-Mounts:


BridgeSDK/BridgeAPI/BridgeAPITypes/MachineGeneratedDoNotEdit/_SBBActivity.h
```diff
@@ -51,10 +51,16 @@
 
 @property (nonatomic, strong) NSString* labelDetail;
```
syoung-smallwisdom: readwrite?

BridgeSDK/BridgeAPI/BridgeAPITypes/MachineGeneratedDoNotEdit/_SBBActivity.m
```diff
@@ -31,15 +31,43 @@
...
 @interface _SBBActivity()
 
 @end
 
+// see xcdoc://?url=developer.apple.com/library/etc/redirect/xcode/ios/602958/documentation/Cocoa/Conceptual/CoreData/Articles/cdAccessorMethods.html
+@interface NSManagedObject (Activity)
+
+@property (nullable, nonatomic, retain) NSString* activityType;
+
+@property (nullable, nonatomic, retain) NSString* guid;
```
syoung-smallwisdom: should the guid be nullable?

BridgeSDK/BridgeAPI/BridgeAPITypes/MachineGeneratedDoNotEdit/_SBBActivity.m
```diff
@@ -31,15 +31,43 @@
...
+#import "SBBTaskReference.h"
+
 @interface _SBBActivity()
 
 @end
 
+// see xcdoc://?url=developer.apple.com/library/etc/redirect/xcode/ios/602958/documentation/Cocoa/Conceptual/CoreData/Articles/cdAccessorMethods.html
+@interface NSManagedObject (Activity)
+
+@property (nullable, nonatomic, retain) NSString* activityType;
```
syoung-smallwisdom: nullable?

BridgeSDK/BridgeAPI/BridgeAPITypes/MachineGeneratedDoNotEdit/_SBBActivity.m
```diff
@@ -89,23 +129,190 @@ - (NSDictionary *)dictionaryRepresentation
...
+    NSManagedObject *relMoTask = [self.task createInContext:cacheContext withObjectManager:objectManager cacheManager:cacheManager];
+
+    [managedObject setTask:relMoTask];
+
+    // Calling code will handle saving these changes to cacheContext.
+}
+
 #pragma mark Direct access
 
+- (void) setSurvey: (SBBSurveyReference*) survey_ settingInverse: (BOOL) setInverse
```
syoung-smallwisdom: what is settingInverse for?

BridgeSDK/BridgeAPI/BridgeAPITypes/MachineGeneratedDoNotEdit/_SBBUserProfile.m
```diff
@@ -31,15 +31,29 @@
...
+
+@property (nullable, nonatomic, retain) NSString* email;
+
+@property (nullable, nonatomic, retain) NSString* firstName;
+
+@property (nullable, nonatomic, retain) NSString* lastName;
+
+@end
+
 @implementation _SBBUserProfile
```
syoung-smallwisdom: Is this ever actually cached? Should it be?

BridgeSDK/BridgeAPI/SBBActivityManager.m
```diff
@@ -56,13 +58,80 @@ + (instancetype)defaultComponent
...
+                if (completion) {
+                    completion(tasks, nil);
+                }
+            });
+            return nil;
+        }
+        
+        // otherwise, keep unfinished tasks from yesterday so we can show them in the Yesterday section
+        // (because the server doesn't give us those)
+        if (daysBehind > 0) {
```
syoung-smallwisdom: Wouldn't we also want to return the finished tasks so that we can show the checkmark if finished?

BridgeSDK/BridgeAPI/BridgeAPITypes/MachineGeneratedDoNotEdit/_SBBActivity.h
Erin-Mounts: *shrug* this is how the template that comes with mogenerator does it, and I only changed what I needed to change. For whatever reason the template adds a superfluous "readwrite" specifier to properties generated from relationships and not to those from attributes.

BridgeSDK/BridgeAPI/SBBAuthManager.m
```diff
@@ -363,6 +363,34 @@ - (NSURLSessionDataTask *)signOutWithCompletion:(SBBNetworkManagerCompletionBloc
     }];
 }
 
+- (NSString *)savedUsername
+{
+    NSString *username = nil;
+    if (_authDelegate) {
+        if ([_authDelegate respondsToSelector:@selector(usernameForAuthManager:)]) {
```
syoung-smallwisdom: why the double nesting? This means that it will only fallback to the usernameFromKeychain if the authDelegate is nil, rather than if the authDelegate does not implement the protocol method?

````
if ([_authDelegate respondsToSelector:@selector(usernameForAuthManager:)]) { 
    return [_authDelegate usernameForAuthManager:self];
} else {
    return [self usernameFromKeychain];
}
````

File: BridgeSDK/BridgeAPI/SBBAuthManager.m [hunk 8]
```diff
@@ -363,6 +363,34 @@ - (NSURLSessionDataTask *)signOutWithCompletion:(SBBNetworkManagerCompletionBloc
...
+    }
+    
+    return username;
+}
+
+- (NSString *)savedPassword
+{
+    NSString *password = nil;
+    if (_authDelegate) {
+        if ([_authDelegate respondsToSelector:@selector(passwordForAuthManager:)]) {
```
syoung-smallwisdom: same question as above.

syoung-smallwisdom:
A few questions. otherwise, looks good.

BridgeSDK/BridgeAPI/BridgeAPITypes/MachineGeneratedDoNotEdit/_SBBActivity.m
Erin-Mounts: this one (and the guid one below) are generated by the template to match what gets generated if you choose Create NSManagedObject Subclass... from the Editor menu when you select an entity in the model editor. Since these declarations are internal to the machine-generated files it shouldn't matter (and I think there are intermediate states e.g. while creating a new instance where the nascent object might not have these values yet, so I don't think I want to make them non-optional in the model, which doesn't seem to affect the generated code anyway)

BridgeSDK/BridgeAPI/BridgeAPITypes/MachineGeneratedDoNotEdit/_SBBActivity.m
Erin-Mounts: This is generated by the template and is meant to handle keeping the inverse relationship connected properly in the PONSO (Plain Old NSObjects) generated code. Since I don't want those inverse relationships to show up in the dictionary/Bridge JSON that gets generated from these objects, I added a flag "notInPONSODictionary" in the userInfo on these relationships, and had it also prevent that code from being generated. We don't use it anyway and the methods with that parameter aren't publicly available.

BridgeSDK/BridgeAPI/BridgeAPITypes/MachineGeneratedDoNotEdit/_SBBUserProfile.m
Erin-Mounts: technically, yes, it's cached automatically by SBBObjectManager when it creates the SBBUserProfile object from Bridge JSON--for all the good it does, since I haven't updated the interfaces to SBBUserManager to make use of the cache. I'll do that.

BridgeSDK/BridgeAPI/SBBActivityManager.m
Erin-Mounts: ah, yes, good point... but only the ones finished today, right?

BridgeSDK/BridgeAPI/SBBAuthManager.m
Erin-Mounts: by design, yes, and documented; see the doc comments for SBBAuthManagerDelegateProtocol and its optional methods in SBBAuthManager.h for details.

BridgeSDK/BridgeAPI/SBBAuthManager.m
Erin-Mounts: same answer ;)

BridgeSDK/BridgeAPI/BridgeAPITypes/MachineGeneratedDoNotEdit/_SBBUserProfile.m
syoung-smallwisdom: But if it's cached in core data that's *not* the keychain even though it's PII?

BridgeSDK/BridgeAPI/SBBActivityManager.m
syoung-smallwisdom: Seems to me that you'd want the app developers to be able to filter on that rather than assume they only want the unfinished ones.

BridgeSDK/BridgeAPI/SBBActivityManager.m
```diff
@@ -56,13 +58,80 @@ + (instancetype)defaultComponent
...
 
-- (NSURLSessionDataTask *)getScheduledActivitiesForDaysAhead:(NSInteger)daysAhead withCompletion:(SBBActivityManagerGetCompletionBlock)completion
+- (NSURLSessionDataTask *)getScheduledActivitiesForDaysAhead:(NSInteger)daysAhead withCompletion:(SBBActivityManagerGetCompletionBlock)completion {
+    SBBCachingPolicy policy = gSBBUseCache ? SBBCachingPolicyFallBackToCached : SBBCachingPolicyNoCaching;
+    return [self getScheduledActivitiesForDaysAhead:daysAhead cachingPolicy:policy withCompletion:completion];
+}
+
+- (NSURLSessionDataTask *)getScheduledActivitiesForDaysAhead:(NSInteger)daysAhead cachingPolicy:(SBBCachingPolicy)policy withCompletion:(SBBActivityManagerGetCompletionBlock)completion
+{
+    return [self getScheduledActivitiesForDaysAhead:daysAhead daysBehind:0 cachingPolicy:policy withCompletion:completion];
```
syoung-smallwisdom: shouldn't this be 1 day by default?

BridgeSDK/BridgeAPI/SBBActivityManager.m
```diff
@@ -56,13 +58,80 @@ + (instancetype)defaultComponent
...
     [self.authManager addAuthHeaderToHeaders:headers];
     
+    __block SBBResourceList *tasks = nil;
+    NSArray *daysBehindTasks = [NSArray array];
+    id<SBBCacheManagerProtocol> cacheManager = nil;
+    if (policy != SBBCachingPolicyNoCaching) {
+        if ([self.objectManager conformsToProtocol:@protocol(SBBObjectManagerInternalProtocol)]) {
+            cacheManager = ((id<SBBObjectManagerInternalProtocol>)self.objectManager).cacheManager;
+        }
+        tasks = (SBBResourceList *)[cacheManager cachedSingletonObjectOfType:@"ResourceList" createIfMissing:NO];
```
syoung-smallwisdom: What happens when the user signs out?

BridgeSDK/BridgeAPI/SBBActivityManager.m
```diff
@@ -56,13 +58,80 @@ + (instancetype)defaultComponent
...
+
+- (NSURLSessionDataTask *)getScheduledActivitiesForDaysAhead:(NSInteger)daysAhead daysBehind:(NSInteger)daysBehind cachingPolicy:(SBBCachingPolicy)policy withCompletion:(SBBActivityManagerGetCompletionBlock)completion
 {
     NSMutableDictionary *headers = [NSMutableDictionary dictionary];
     [self.authManager addAuthHeaderToHeaders:headers];
     
+    __block SBBResourceList *tasks = nil;
+    NSArray *daysBehindTasks = [NSArray array];
+    id<SBBCacheManagerProtocol> cacheManager = nil;
+    if (policy != SBBCachingPolicyNoCaching) {
```
syoung-smallwisdom: This will result in an unexpected result for the developer who wants to say "Always ask the server" and expects the server to be able to return the tasks for days in the past.

BridgeSDK/BridgeAPI/SBBActivityManager.m
```diff
@@ -56,13 +58,80 @@ + (instancetype)defaultComponent
...
+            
+            // ...ok, if we *did* update from the server though, we want to add back the ones left over from
+            // the daysBehind window.
+            if (!error && daysBehindTasks.count) {
+                [tasks insertItems:daysBehindTasks atIndexes:[NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, daysBehindTasks.count)]];
+                
+                // ...and save it back to the cache for later.
+                [tasks saveToCoreDataCacheWithObjectManager:self.objectManager];
+            }
+        }
```
syoung-smallwisdom: This is confusing. Why are you using `__block` for `tasks` when you are setting it three different times?  

Additionally, you've got a bug in the code to add the `daysBehindTask` back in. So... developer does a background grab of *only* tasks that need notifications set up for them (tasks that are in the future and do not yet have notifications set up) so `daysBehind=0` then they want to show the users activity where they show what the user has done already so `daysBehind=1`, but the cache will at this point only contain days ahead b/c the days behind have been culled by the previous call.

BridgeSDK/BridgeAPI/BridgeAPITypes/MachineGeneratedDoNotEdit/_SBBUserProfile.m
Erin-Mounts: Oh yeah. I'll mark its model with the "encrypted" flag in userData so the template will generate code to encrypt/decrypt when writing to/reading from persistent store.

BridgeSDK/BridgeAPI/SBBActivityManager.m
Erin-Mounts: Why using __block for tasks? Because at one point in the evolution of this code it was necessary/made sense and then I never noticed to go back and fix it when it didn't anymore.

As for the bug, this is what I get for not thoroughly unit testing *before* submitting PRs. I think the solution for this (and the previous comment about including all, not just unfinished, "daysBehind" activities) is to just always save X days worth of cached activities from the past (and any marked as finished, whether in the past or not, because the server won't send those anymore either) and add those back to the cached ResourceList items after fetching from the server, and then filter and send just the requested time span to the completion handler. The question then is what's a good value for X days (i.e. a max value for daysBehind). The longest any of our apps currently need is 7 days; should I go with that for now, do you think? Or should I add it as a configuration parameter in the SDK setup call?

BridgeSDK/BridgeAPI/SBBActivityManager.m
Erin-Mounts: I was thinking it didn't have to be backward-compatible since anyone using BridgeSDK currently is handling this on their own (e.g. in AppCore), so if they refactored to take advantage of caching they'd explicitly state how many daysBehind they want anyway. But yeah, if the default is 1 most AppCore apps wouldn't actually have to update this call and they'd still get what they need, so that's better. I'll change it.

BridgeSDK/BridgeAPI/SBBActivityManager.m
Erin-Mounts: Hm, yeah. I was thinking of "NoCaching" policy here as being equivalent to not having caching turned on, but it's not.

BridgeSDK/BridgeAPI/BridgeAPITypes/MachineGeneratedDoNotEdit/_SBBUserProfile.m
syoung-smallwisdom: cool, thanks.

BridgeSDK/BridgeAPI/SBBActivityManager.m
Erin-Mounts: Hm, this is a bit tricky. On the one hand, to make sure there's no interference if they sign in with a different account after signing out, we could clear the cache when they sign out. If we do that though, then when they signed back in to the same account they'd have lost all history of expired and finished scheduled activities.

I think the way to handle this is to not flush/clear the cache when they sign out, but instead to add a hash of the email address they signed in with to the persistent store path, so they'll get a separate cache for each email they sign in with, and if they sign in with the same one they'll get their old cache back.

BridgeSDK/BridgeAPI/SBBActivityManager.m
syoung-smallwisdom: Let's just go with 7 days for now. This can always be revised in the future, but there are so many changes in this pull request that github is already overloaded.

BridgeSDK/BridgeAPI/SBBActivityManager.m
syoung-smallwisdom: Yup, Is there a link in the model between the resource list and the current user? Do we know who that user is (via session)? The whole concept of a singleton is assuming that the app does not cache data once the user has logged out. Ideally, this needs to be recoverable via the server, though that is a backlog task for another day. Since any current design does not require saving state for a logged out user, I think it would be fine to flush the cache. Recommend via cascade delete and/or batch delete verse deleting the persistent store sqlite file (persistent store file delete is faster but brittle).

BridgeSDK/BridgeAPI/SBBActivityManager.m
```diff
@@ -65,75 +67,99 @@ - (NSURLSessionDataTask *)getScheduledActivitiesForDaysAhead:(NSInteger)daysAhea
...
+        BOOL isInternalObjectManager = [self.objectManager conformsToProtocol:@protocol(SBBObjectManagerInternalProtocol)];
+        if (isInternalObjectManager) {
             cacheManager = ((id<SBBObjectManagerInternalProtocol>)self.objectManager).cacheManager;
         }
-        tasks = (SBBResourceList *)[cacheManager cachedSingletonObjectOfType:@"ResourceList" createIfMissing:NO];
+        SBBResourceList *tasks = (SBBResourceList *)[cacheManager cachedSingletonObjectOfType:@"ResourceList" createIfMissing:NO];
         
         // if we're going straight to cache, we're done
         if (policy == SBBCachingPolicyCachedOnly) {
             dispatch_async(dispatch_get_main_queue(), ^{
```
syoung-smallwisdom: why is this dispatched on the main queue?

BridgeSDK/BridgeAPI/SBBActivityManager.m
```diff
@@ -65,75 +67,99 @@ - (NSURLSessionDataTask *)getScheduledActivitiesForDaysAhead:(NSInteger)daysAhea
...
-            }];
-            
-            daysBehindTasks = [tasks.items filteredArrayUsingPredicate:predicate];
-        }
+        savedTasks = [self filterTasks:tasks.items forDaysAhead:kMaxAdvance andDaysBehind:kDaysToCache excludeStillValid:YES];
     }
 
     return [self.networkManager get:kSBBActivityAPI headers:headers parameters:@{@"daysAhead": @(daysAhead), @"offset": [[NSDate date] ISO8601OffsetString]} completion:^(NSURLSessionDataTask *task, id responseObject, NSError *error) {
-        tasks = [self.objectManager objectFromBridgeJSON:responseObject];
+        SBBResourceList *tasks = [self.objectManager objectFromBridgeJSON:responseObject];
```
syoung-smallwisdom: why pull the object from the bridge JSON here when replacing with cached version in the step below?

````
SBBResourceList *tasks = nil;
if (policy != SBBCachingPolicyFallBackToCached) {
    tasks = [self.objectManager objectFromBridgeJSON:responseObject];
}
else {
...
````

File: BridgeSDK/BridgeAPI/SBBActivityManager.h [hunk 15]
```diff
@@ -37,10 +37,10 @@
 /*!
  Completion block called when retrieving scheduled activities from the API.
  
- @param activitiesList By default, an SBBResourceList object, unless the ResourceList type has been mapped in SBBObjectManager setupMappingForType:toClass:fieldToPropertyMappings:. The item property (or whatever it was mapped to) contains an NSArray of SBBScheduledActivity objects and the total (or mapped-to) property contains an NSNumber indicating how many ScheduledActivity objects were retrieved--again, unless the ScheduledActivity type has been mapped to a different class.
+ @param activitiesList By default, an array of SBBScheduledActivity objects, unless the ScheduledActivity type has been mapped in SBBObjectManager setupMappingForType:toClass:fieldToPropertyMappings:.
  @param error       An error that occurred during execution of the method for which this is a completion block, or nil.
  */
-typedef void (^SBBActivityManagerGetCompletionBlock)(id activitiesList, NSError *error);
+typedef void (^SBBActivityManagerGetCompletionBlock)(NSArray *activitiesList, NSError *error);
```
syoung-smallwisdom: Won't changing the returned object cause AppCore implementation to fail?

BridgeSDK/BridgeAPI/SBBActivityManager.m
```diff
@@ -65,75 +67,99 @@ - (NSURLSessionDataTask *)getScheduledActivitiesForDaysAhead:(NSInteger)daysAhea
...
         }
         if (completion) {
-            completion(tasks, error);
+            NSMutableArray *requestedTasks = [[self filterTasks:tasks.items forDaysAhead:daysAhead andDaysBehind:daysBehind excludeStillValid:NO] mutableCopy];
+            if ([self.objectManager conformsToProtocol:@protocol(SBBObjectManagerInternalProtocol)]) {
+                // in case object mapping has been set up
+                for (NSInteger i = 0; i < requestedTasks.count; ++i) {
+                    requestedTasks[i] = [((id<SBBObjectManagerInternalProtocol>)self.objectManager) mappedObjectForBridgeObject:requestedTasks[i]];
+                }
+            }
```
syoung-smallwisdom: can you pull this out as a block or a convenience method? Looks very similar to the code above.

syoung-smallwisdom:
Few more questions.

BridgeSDK/BridgeAPI/SBBActivityManager.m
Erin-Mounts: I'm... not entirely sure what I was thinking here. If I wanted to just get it off onto another queue so the caller didn't block while the handler ran, then the main queue is probably exactly the *wrong* queue to do it in. The "normal" call to this method's handler would be called from the network request completion handler, which is documented to run on the "delegate queue" for the NSURLSession but since it's the default session and you can't set a delegate on it, I'm not entirely sure what that would mean, or if it would even *have* a delegate queue. *Shrug.* I'll take it out; it doesn't really seem necessary.

BridgeSDK/BridgeAPI/SBBActivityManager.m
Erin-Mounts: The objectFromBridgeJSON call is how objects get cached or have their cached representations updated; that's the reason I call it here. Although always cached as Bridge objects, what that method returns may potentially have been mapped to a different object type, if mappings have been set up in this instance of the object manager. So below, once I know the cached version is current, I pull it directly from the cache to make sure I always am working with genuine SBB objects while I'm processing everything. Then below I call the mapping routine to make sure I'm passing the expected objects to the completion handler.

Since no one is actually using the mapping feature AFAIK, it might be worth considering just pulling it out to simplify maintenance going forward, but that would be a job for its own separate pull request.

BridgeSDK/BridgeAPI/SBBActivityManager.m
Erin-Mounts: Good idea.

BridgeSDK/BridgeAPI/SBBActivityManager.h
Erin-Mounts: Only if someone using AppCore decides to upgrade to this version of BridgeSDK, and then it's trivial to update (and trivial to notice since the app calls this method every time the Activity screen gets foregrounded or refreshed). Doing it this way lets me keep one canonical SBBResourceList object per item type, enforced by the cache and object managers, while handing back arbitrary subsets of its items to callers, and there's no information in an SBBResourceList that isn't inherent in the array itself so they don't lose anything.

syoung-smallwisdom: But this is used as a drop through by AppCore. 

````
            [SBBComponent (SBBActivityManager) getScheduledActivitiesForDaysAhead:4
                                          withCompletion:^(SBBResourceList *tasksList,
                                                           NSError *errorFetchingTasks)
````
(copied from APCScheduler)

PR #69, opened 2016-04-28
Fix assignment to self in initWithManagedObject:

Erin-Mounts:
fixed in template and regenerated object code. cleared up analyzer warning.


syoung-smallwisdom:
thanks

PR #72, opened 2016-04-28
Fix md5-to-hex-string bug

Erin-Mounts:
unsigned char so if the high bit of a character is set it doesn't left-pad with FFF....


syoung-smallwisdom:
looks good

PR #73, opened 2016-04-29
Fix template

Erin-Mounts:
so it doesn't append the same relationship items to the existing list every time it updates an object

### DIFF
--- a/BridgeSDK/BridgeAPI/SBBActivityManager.m
+++ b/BridgeSDK/BridgeAPI/SBBActivityManager.m
@@ -79,9 +79,11 @@ NSInteger const     kMaxAdvance  =       4; // server only supports 4 days ahead
     NSDate *windowEnd = [cal startOfDayForDate:[todayEnd dateByAddingTimeInterval:daysAhead * kSBB24Hours]];
     
     // things that either expired during the daysBefore period, or expire after that but start before the end of daysAhead
-    NSPredicate *predicate = [NSPredicate predicateWithFormat:@"(%K > %@ AND %K < %@) OR (%K >= %@ AND %K < %@)",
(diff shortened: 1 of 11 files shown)


BridgeSDK/BridgeAPI/SBBActivityManager.m
```diff
@@ -138,7 +140,8 @@ - (NSURLSessionDataTask *)getScheduledActivitiesForDaysAhead:(NSInteger)daysAhea
         savedTasks = [self filterTasks:tasks.items forDaysAhead:kMaxAdvance andDaysBehind:kDaysToCache excludeStillValid:YES];
     }
 
-    return [self.networkManager get:kSBBActivityAPI headers:headers parameters:@{@"daysAhead": @(daysAhead), @"offset": [[NSDate date] ISO8601OffsetString]} completion:^(NSURLSessionDataTask *task, id responseObject, NSError *error) {
+    // always request the maximum days ahead from the server so we have them cached
+    return [self.networkManager get:kSBBActivityAPI headers:headers parameters:@{@"daysAhead": @(kMaxAdvance), @"offset": [[NSDate date] ISO8601OffsetString]} completion:^(NSURLSessionDataTask *task, id responseObject, NSError *error) {
```
syoung-smallwisdom: How will caching work for when the schedules change (b/c of changes to data groups)? This still does a replacement, right? (i.e. server wins)

Erin-Mounts: server "wins", but we cheat on this one: before updating, SBBActivityManager saves everything we want to keep that the server would leave out/overwrite/delete (i.e. everything going back 1 week in the past and everything from today on that's been marked finished), and then adds it back in after updating from server.

syoung-smallwisdom: cool, thanks

syoung-smallwisdom:
looks good

PR #76, opened 2016-05-17
BRIDGE-1343 Server only returns items that are NOT expired and NOT complete

syoung-smallwisdom:
Fixed bug where "expired" was defined as "start of today" whereas on the server, expired is defined as "now".

### DIFF
--- a/BridgeSDK/BridgeAPI/SBBActivityManager.m
+++ b/BridgeSDK/BridgeAPI/SBBActivityManager.m
@@ -89,6 +89,7 @@ NSInteger const     kMaxAdvance  =       4; // server only supports 4 days ahead
     NSCalendar *cal = [NSCalendar currentCalendar];
     NSDate *windowStart = [cal startOfDayForDate:[NSDate dateWithTimeIntervalSinceNow:-daysBehind * kSBB24Hours]];
     NSDate *todayStart = [cal startOfDayForDate:[NSDate date]];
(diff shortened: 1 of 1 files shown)


BridgeSDK/BridgeAPI/SBBActivityManager.m
```diff
@@ -103,11 +104,11 @@ - (NSArray *)filterTasks:(NSArray *)tasks forDaysAhead:(NSInteger)daysAhead andD
     NSArray *filtered = [tasks filteredArrayUsingPredicate:predicate];
     if (excludeValid) {
         // "valid things" are defined as the subset of the above things that the server would return.
-        // valid things expire after the start of today and are not marked as finished. the flag is
+        // valid things expire after now and are not marked as finished. the flag is
         // set so we want to exclude those things.
-        predicate = [NSPredicate predicateWithFormat:@"NOT ((%K == nil OR %K > %@) AND %K == nil)",
+        predicate = [NSPredicate predicateWithFormat:@"NOT ((%K == nil OR %K >= %@) AND %K == nil)",
```
Erin-Mounts: not sure whether '>' or '>=' is more correct in terms of matching server behavior, but since the odds of this code using precisely the same time instant the server used for checking against are infinitesimal, it shouldn't really matter either way.

Erin-Mounts:
Looks good (one irrelevant comment)

PR #78, opened 2016-06-03
FPHS 561

EricSiegNW:
Prevent caching of signup/signin POST requests that include the password.

Basically, iOS really likes to cache things that use NSURLRequest. While Bridge server is careful to not return the password in any responses, iOS will still cache POSTs that we make to the server. The POST requests that appear to have relevant info (password especially) are SignIn and SignUp. To prevent these POSTs from being cached, we need to add a cache-control header to the request.

There are a few different approaches, but this seemed to be the simplest and best and appeared to work after multiple before and after tests.

The only other call that appeared to have a risk of having this same issue was resetPasswordToNewPassword so the change was applied there as well although I haven't done before/after tests. 

Repro steps to access cached password before this change is applied:

Launch app
sign in
make it to main screen
download the cache.db-wal file and open it in xcode, then search for your password


Erin-Mounts:
looks good

PR #79, opened 2016-06-08
Bridge 1382

Erin-Mounts:
- Allow any Bridge request to be done with the background session (so it can be initiated offline) by adding a BOOL "background" parameter to the basic networking methods.
- Refactor interfaces to use NSURLSessionTask instead of NSURLSessionDataTask:
--it's the parent class of both NSURLSessionDataTask and NSURLSessionDownloadTask, so the same set of methods can be used for both with no change other than the background parameter
--it doesn't affect any calling code beyond changing the class name in some completion handler parameter lists, because NSURLSessionDataTask doesn't add any properties or methods to the parent (abstract) class
--I should have done it this way in the first place
- Have requests using the background session to access Bridge APIs respond to HTTP errors (401, 410, etc.) the same as foreground requests
- Refactor the uploadFileToBridge:etc. code to take advantage of this, and to use completion handlers in place of delegate methods
- Update SBBActivityManager updateScheduledActivities:withCompletion: to use the background session


BridgeSDK/BridgeAPI/SBBBridgeNetworkManager.m
```diff
@@ -74,11 +74,12 @@ - (instancetype)initWithAuthManager:(id<SBBAuthManagerInternalProtocol>)authMana
     return self;
 }
 
-- (NSURLSessionDataTask *) doDataTask: (NSString*) method
-                            URLString: (NSString*)URLString
-                              headers: (NSDictionary *)headers
-                           parameters:(NSDictionary *)parameters
-                           completion:(SBBNetworkManagerCompletionBlock)completion
+- (NSURLSessionTask *)doDataTask:(NSString *)method
```
syoung-smallwisdom: Is this intended as an internal method only? (ie. should the old version be deprecated?)

BridgeSDK/BridgeAPI/SBBUploadManager.m
```diff
@@ -255,6 +255,125 @@ - (SBBUploadSession *)uploadSessionForFile:(NSString *)fileURLString
     return uploadSession;
 }
 
+// When uploading a file to Bridge, first we use background download to get an SBBUploadSession object from Bridge.
+// Then we use the pre-signed URL in that SBBUploadSession to upload our file to S3.
+// When that finishes, we tell Bridge that the requested upload is completed.
+// To avoid having multiple layers of nested completion handlers in one method, they're broken out by step:
+
+// --- Here's the completion handler for the step where we tell Bridge the upload completed:
+- (void)toldBridgeFileUploadedWithTask:(NSURLSessionUploadTask *)uploadTask forBridgeSession:(SBBUploadSession *)uploadSession error:(NSError *)error
```
syoung-smallwisdom: nit: I prefer "handle" to "told"... or is the meaning something different?

BridgeSDK/BridgeAPI/SBBBridgeNetworkManager.m
Erin-Mounts: it's internal-only

BridgeSDK/BridgeAPI/SBBUploadManager.m
```diff
@@ -255,6 +255,125 @@ - (SBBUploadSession *)uploadSessionForFile:(NSString *)fileURLString
...
+    if (httpStatusCode >= 300) {
+        // iOS handles redirects automatically so only e.g. 307 resource not changed etc. from the 300 range should end up here
+        // (along with all 4xx and 5xx of course)
+        NSString *description = [NSString stringWithFormat:@"Background file upload to S3 failed with HTTP status %ld", (long)httpStatusCode];
+        NSError *s3Error = [NSError errorWithDomain:SBB_ERROR_DOMAIN code:SBBErrorCodeS3UploadErrorResponse userInfo:@{NSLocalizedDescriptionKey: description}];
+        [self completeUploadOfFile:uploadTask.taskDescription withError:s3Error];
+        return;
+    }
+    
+    // tell the API we done did it
```
syoung-smallwisdom: "done did it"?

BridgeSDK/BridgeAPI/SBBUploadManager.m
Erin-Mounts: well, it's a "handler" that gets called when we've "told" Bridge the file upload completed...

BridgeSDK/Networking/SBBNetworkManagerInternal.h
```diff
@@ -53,9 +53,16 @@ extern NSString *kAPIPrefix;
 
 - (instancetype)initWithBaseURL:(NSString*)baseURL bridgeStudy:(NSString*)bridgeStudy;
 
-- (void)handleHTTPError:(NSError *)error task:(NSURLSessionDataTask *)task response:(id)responseObject retryObject:(APCNetworkRetryObject *)retryObject;
+- (void)handleHTTPError:(NSError *)error task:(NSURLSessionTask *)task response:(id)responseObject retryObject:(SBBNetworkRetryObject *)retryObject;
 - (NSDictionary *)headersPreparedForRetry:(NSDictionary *)headers;
 
 - (NSURL *) URLForRelativeorAbsoluteURLString: (NSString*) URLString;
 
+- (NSURLSessionTask *) doDataTask:(NSString*) method
```
syoung-smallwisdom: question answered.

BridgeSDK/BridgeAPI/SBBUploadManager.m
Erin-Mounts: moved this code here from where it used to be, mostly unchanged (including this comment)--probably had been reading or watching a lot of something with Southern or country dialect at the time? dunno.

BridgeSDK/BridgeAPI/SBBUploadManager.m
```diff
@@ -255,6 +255,125 @@ - (SBBUploadSession *)uploadSessionForFile:(NSString *)fileURLString
...
+    NSString *uploadId = uploadSession.id;
+    if (!uploadId.length) {
+        NSAssert(uploadId, @"uploadId is nil");
+    }
+    NSString *ref = [NSString stringWithFormat:kSBBUploadCompleteAPIFormat, uploadId];
+    NSMutableDictionary *headers = [NSMutableDictionary dictionary];
+    [self.authManager addAuthHeaderToHeaders:headers];
+    [self.networkManager post:ref headers:headers parameters:nil background:YES completion:^(NSURLSessionTask *task, id responseObject, NSError *error) {
+        [self toldBridgeFileUploadedWithTask:uploadTask forBridgeSession:uploadSession error:error];
+    }];
```
Erin-Mounts: not seeing what you're seeing here...?

syoung-smallwisdom:
looks good

PR #80, opened 2016-06-23
Bridge 1390

Erin-Mounts:
- Handle 403 (presigned url expired) status code from S3 upload by retrying with a fresh presigned url.
- Handle download task errors that come with resume data by using that data to resume the task.

### DIFF
--- a/BridgeSDK/BridgeAPI/SBBUploadManager.m
+++ b/BridgeSDK/BridgeAPI/SBBUploadManager.m
@@ -216,10 +216,15 @@ static NSString *kUploadSessionsKey = @"SBBUploadSessionsKey";
     [defaults synchronize];
 }
 
+- (NSDictionary *)uploadRequestJSONForFile:(NSString *)fileURLString
+{
+    return [[NSUserDefaults standardUserDefaults] dictionaryForKey:kUploadRequestsKey][fileURLString];
+}
+
 - (SBBUploadRequest *)uploadRequestForFile:(NSString *)fileURLString
 {
     SBBUploadRequest *uploadRequest = nil;
-    id json = [[NSUserDefaults standardUserDefaults] dictionaryForKey:kUploadRequestsKey][fileURLString];
+    id json = [self uploadRequestJSONForFile:fileURLString];
     if (json) {
         uploadRequest = [_cleanObjectManager objectFromBridgeJSON:json];
     }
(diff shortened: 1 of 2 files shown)


BridgeSDK/Networking/SBBNetworkManager.m
```diff
@@ -723,6 +723,26 @@ - (void)URLSession:(NSURLSession *)session downloadTask:(NSURLSessionDownloadTas
...
+            resumeTask.taskDescription = task.taskDescription;
+            SBBNetworkManagerTaskCompletionBlock taskCompletion = [self completionBlockForTask:dlTask];
+            SBBNetworkManagerDownloadCompletionBlock downloadCompletion = [self completionBlockForDownload:dlTask];
+            [self setCompletionBlock:taskCompletion forTask:resumeTask];
+            [self setCompletionBlock:downloadCompletion forDownload:resumeTask];
+            [self removeCompletionBlockForTask:dlTask];
+            [self removeCompletionBlockForDownload:dlTask];
+            
+            [resumeTask resume];
+            return;
```
syoung-smallwisdom: This is the kind of early exit where I really like the Swift `guard`. Could you extract this into a method to make the early exit a little more obvious? I worry that it could get lost in a merge or other kind of "cleanup".

BridgeSDK/BridgeAPI/SBBUploadManager.m
```diff
@@ -360,7 +375,7 @@ - (void)downloadedBridgeUploadSessionWithDownloadTask:(NSURLSessionDownloadTask
         NSURL *fileUrl = [NSURL fileURLWithPath:downloadTask.taskDescription];
         [self.networkManager uploadFile:fileUrl httpHeaders:uploadHeaders toUrl:uploadSession.url taskDescription:downloadTask.taskDescription completion:^(NSURLSessionTask *task, NSHTTPURLResponse *response, NSError *error) {
 #if DEBUG
-            if (error) {
+            if (error || response.statusCode >= 300) {
                 NSLog(@"Error uploading to S3 for upload ID %@:\n%@", uploadSession.id, error);
```
syoung-smallwisdom: Include the statusCode in the debug log.

BridgeSDK/BridgeAPI/SBBUploadManager.m
```diff
@@ -286,26 +291,36 @@ - (void)toldBridgeFileUploadedWithTask:(NSURLSessionUploadTask *)uploadTask forB
...
+        return;
+    } else if (httpStatusCode >= 300) {
         // iOS handles redirects automatically so only e.g. 307 resource not changed etc. from the 300 range should end up here
         // (along with all 4xx and 5xx of course)
         NSString *description = [NSString stringWithFormat:@"Background file upload to S3 failed with HTTP status %ld", (long)httpStatusCode];
         NSError *s3Error = [NSError errorWithDomain:SBB_ERROR_DOMAIN code:SBBErrorCodeS3UploadErrorResponse userInfo:@{NSLocalizedDescriptionKey: description}];
-        [self completeUploadOfFile:uploadTask.taskDescription withError:s3Error];
+        [self completeUploadOfFile:filePath withError:s3Error];
         return;
     }
```
syoung-smallwisdom: This is the kind of a case where I find nested if statements and explicit special-case handling a little easier to read.

````
    // server didn't like the request, or otherwise hiccupped
    if (httpStatusCode >= 300) {
        switch (httpStatusCode) {
            case 403:{
#if DEBUG
                NSLog(@"Background file upload to S3 failed with HTTP status 403 (presumably due to presigned url timeout); retrying...");
#endif
                // this will have happened because the pre-signed url timed out before starting the actual upload to S3,
                // so try, try again
                NSDictionary *uploadRequestJSON = [self uploadRequestJSONForFile:filePath];
                [self kickOffUploadForFile:filePath uploadRequestJSON:uploadRequestJSON];
            } break;
                
            default:{
                // iOS handles redirects automatically so only e.g. 307 resource not changed etc. from the 300 range should end up here
                // (along with all 4xx and 5xx of course)
                NSString *description = [NSString stringWithFormat:@"Background file upload to S3 failed with HTTP status %ld", (long)httpStatusCode];
                NSError *s3Error = [NSError errorWithDomain:SBB_ERROR_DOMAIN code:SBBErrorCodeS3UploadErrorResponse userInfo:@{NSLocalizedDescriptionKey: description}];
                [self completeUploadOfFile:filePath withError:s3Error];
            } break;
        }

        // exit early
        return;
    }
````

Possibly moving the entire switch statement into an extracted method for readability?

syoung-smallwisdom:
Few suggestions for code rearranging to improve early exit readability. Otherwise, looks good.

syoung-smallwisdom:
lovely.

PR #85, opened 2016-07-18
bug/FPHS-690

dephillipsmichael:
After a user withdraws, and then they try to sign in again, app was crashing because consent status was nil. 

Once I added a nil check, everything worked again with the consent flow after withdraw.


syoung-smallwisdom:
You are trying to change the machine-generated code.

dephillipsmichael:
Whoops, ok.  I'll have to look into this deeper and see how to change it before the machine generated code.

PR #86, opened 2016-07-29
Unit testing for uploads, and fixes to problems exposed thereby

Erin-Mounts:


BridgeSDK/BridgeAPI/SBBUploadManager.m
```diff
@@ -46,14 +46,16 @@
...
-static NSString * const kSBBUploadAPI =                 UPLOAD_API;
-static NSString * const kSBBUploadCompleteAPIFormat =   UPLOAD_API @"/%@/complete";
-static NSString * const kSBBUploadStatusAPIFormat =     UPLOAD_STATUS_API @"/%@";
+NSString * const kSBBUploadAPI =                 UPLOAD_API;
+NSString * const kSBBUploadCompleteAPIFormat =   UPLOAD_API @"/%@/complete";
+NSString * const kSBBUploadStatusAPIFormat =     UPLOAD_STATUS_API @"/%@";
 
 static NSString *kUploadFilesKey = @"SBBUploadFilesKey";
 static NSString *kUploadRequestsKey = @"SBBUploadRequestsKey";
 static NSString *kUploadSessionsKey = @"SBBUploadSessionsKey";
```
syoung-smallwisdom: Is there a reason for not changing these as well?

Erin-Mounts: Only needed the API strings to be extern so they'd be available for the unit tests.

BridgeSDK/BridgeAPI/SBBUploadManager.m
```diff
@@ -427,43 +422,51 @@ - (void)uploadedFileToS3WithTask:(NSURLSessionUploadTask *)uploadTask response:(
 - (void)downloadedBridgeUploadSessionWithDownloadTask:(NSURLSessionDownloadTask *)downloadTask fileURL:(NSURL *)file
 {
     NSError *error;
+    NSString *uploadFileURL = downloadTask.taskDescription;
```
syoung-smallwisdom: copy? not sure if the taskDescription pointer can be nil'ed and released?

Erin-Mounts: not sure I get what you're concerned about here--if taskDescription did somehow get nil'ed, ARC would ensure it doesn't get released immediately because of this assignment, no?

BridgeSDK/BridgeAPI/SBBUploadManager.m
```diff
@@ -544,42 +547,109 @@ - (void)uploadFileToBridge:(NSURL *)fileUrl contentType:(NSString *)contentType
 - (void)kickOffUploadForFile:(NSString *)filePath uploadRequestJSON:(NSDictionary *)uploadRequestJSON
 {
     // this starts the process by downloading the Bridge UploadSession.
+    // first, make sure we get rid of any existing UploadSession for this file, in case this is a retry.
+    [self setUploadSessionJSON:nil forFile:filePath];
```
syoung-smallwisdom: Not sure I understand the purpose of this. why remove the object from the dictionary rather than just replace with the new key? Wouldn't it be simpler and safer to just `@synchronize` the method `setUploadRequestJSON:forFile:`?

BridgeSDK/BridgeAPI/SBBUploadManagerInternal.h
```diff
@@ -0,0 +1,50 @@
+//
```
syoung-smallwisdom: Nit: Should this be called `SBBUploadManager+Internal.h`?

BridgeSDK/Networking/SBBNetworkManager.m
```diff
@@ -696,6 +696,17 @@ - (NSURL *) URLForRelativeorAbsoluteURLString: (NSString*) URLString
...
 }
 
+- (void)performBlockOnBackgroundDelegateQueue:(void (^)(void))block
+{
+    NSOperationQueue *bgQueue = _backgroundSession.delegateQueue;
+    if (bgQueue) {
+        [bgQueue addOperationWithBlock:block];
+    } else {
+        // do it in line
+        block();
```
syoung-smallwisdom: should this be dispatched async?

BridgeSDK/BridgeAPI/SBBUploadManager.m
syoung-smallwisdom: oh yeah. sorry.

syoung-smallwisdom:
Few nit questions. Otherwise, looks good. Feel free to merge.

BridgeSDK/BridgeAPI/SBBUploadManager.m
Erin-Mounts: it's effectively synchronized by only being called from within the background session delegate queue. I remove the upload session here so the completion handler doesn't see it's already got one and ignore it as a duplicate session due to the retry crossing paths with the response finally coming in from the original request that was being retried.

BridgeSDK/BridgeAPI/SBBUploadManagerInternal.h
Erin-Mounts: it's not defining the category SBBUploadManager(Internal), it's just a "private" header to give unit tests access to some internal stuff.

PR #90, opened 2016-08-22
Fix file path handling

Erin-Mounts:


BridgeSDK/BridgeAPI/SBBUploadManager.m
```diff
@@ -134,23 +201,39 @@ - (NSURL *)tempUploadDirURL
...
+    fileURL = [fileURL URLByResolvingSymlinksInPath];
+    
+    // don't stack UUIDs in front of the base filename--replace any number of them with one new one; one UUID is enough to guarantee uniqueness
+    NSString *baseFileName = [fileURL lastPathComponent];
+    NSRegularExpression *regex = [NSRegularExpression regularExpressionWithPattern:uuidPrefixRegexPattern options:0 error:nil];
+    NSTextCheckingResult *match;
+    while ((match = [regex firstMatchInString:baseFileName options:0 range:NSMakeRange(0, baseFileName.length)])) {
+        baseFileName = [baseFileName substringFromIndex:match.range.length];
+    }
+    NSString *fileName = [NSString stringWithFormat:@"%@_%@", [[NSUUID UUID] UUIDString], baseFileName];
```
syoung-smallwisdom: Why does this need to be here rather than a single check "if the filename already includes a UUID then don't add one"?

syoung-smallwisdom:
looks ok

PR #92, opened 2016-09-09
BRIDGE-1474 Setup hooks for allowing more than 4 days caching ahead and 7 days behind

syoung-smallwisdom:
This *only* addresses getting more schedules from the server (and caching them) and does not handle caching activity scores.

Required for Smart4SURE app scheduling to get more days out.

### DIFF
--- a/BridgeSDK/BridgeAPI/SBBActivityManager.m
+++ b/BridgeSDK/BridgeAPI/SBBActivityManager.m
@@ -181,7 +181,7 @@ NSInteger const     kMaxAdvance  =       4; // server only supports 4 days ahead
     [self.authManager addAuthHeaderToHeaders:headers];
     
     // If caching, always request the maximum days ahead from the server so we have them cached
-    // and if the desiredDaysAhead is less that the max allowed, then fetch a minimum number of
+    // and if the desiredDaysAhead is more than the max allowed, then fetch a minimum number of
     // items per schedule.
     NSInteger desiredDaysAhead = gSBBUseCache ? gSBBCacheDaysAhead : daysAhead;
     NSInteger fetchDaysAhead = MIN(kMaxAdvance, desiredDaysAhead);
--- a/BridgeSDK/BridgeAPI/SBBActivityManager.m
+++ b/BridgeSDK/BridgeAPI/SBBActivityManager.m
@@ -44,7 +44,6 @@
 
 NSString * const kSBBActivityAPI =       ACTIVITY_API;
 NSTimeInterval const kSBB24Hours =       86400;
-NSInteger const     kDaysToCache =       7;
 NSInteger const     kMaxAdvance  =       4; // server only supports 4 days ahead
 
 @interface SBBActivityManager()<SBBActivityManagerInternalProtocol>
@@ -146,7 +145,7 @@ NSInteger const     kMaxAdvance  =       4; // server only supports 4 days ahead
 
 - (NSArray<SBBScheduledActivity *> *)savedTasksFromCachedTasks:(NSArray<SBBScheduledActivity *> *)cachedTasks
 {
-    return [self filterTasks:cachedTasks forDaysAhead:kMaxAdvance andDaysBehind:kDaysToCache excludeStillValid:YES];
+    return [self filterTasks:cachedTasks forDaysAhead:gSBBCacheDaysAhead andDaysBehind:gSBBCacheDaysBehind excludeStillValid:YES];
 }
 
 - (void)addSavedTasks:(NSArray<SBBScheduledActivity *> *)savedTasks toResourceList:(SBBResourceList *)resourceList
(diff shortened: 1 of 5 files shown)


BridgeSDK/BridgeSDK.h
```diff
@@ -90,6 +97,22 @@ static SBBEnvironment gDefaultEnvironment = kDefaultEnvironment;
 + (void)setupWithStudy:(NSString *)study useCache:(BOOL)useCache environment:(SBBEnvironment)environment;
 
 /*!
+ * Set up the Bridge SDK for the given study and server environment. Usually you would only call this version
```
Erin-Mounts: Re: this code comment only calling this version for testing, there should be another method with the same signature but without the environment parameter, which calls this one with the default environment (see the other setupWithStudy method pairs), which is what would be called from the app to set up Bridge with non-default cache timespans.

BridgeSDK/BridgeSDK.m
```diff
@@ -61,6 +65,20 @@ + (void)setupWithStudy:(NSString *)study useCache:(BOOL)useCache environment:(SB
...
+    // make sure the Bridge network manager is set up as the delegate for the background session
+    [SBBComponent(SBBBridgeNetworkManager) restoreBackgroundSession:kBackgroundSessionIdentifier completionHandler:nil];
+}
+
++ (void)setupWithStudy:(NSString *)study cacheDaysAhead:(NSInteger)cacheDaysAhead cacheDaysBehind:(NSInteger)cacheDaysBehind environment:(SBBEnvironment)environment
+{
+    gSBBAppStudy = study;
+    gSBBDefaultEnvironment = environment;
+    gSBBUseCache = cacheDaysAhead > 0 || cacheDaysBehind > 0;
+    gSBBCacheDaysAhead = MIN(SBBMaxSupportedCacheDays, cacheDaysAhead);
```
Erin-Mounts: SBBMaxSupportedCacheDays is 30, but in SBBActivityManager, kMaxAdvance will cap this at 4 days. Is the 4 day restriction on what the server returns still valid? Depending on the answer, one value or the other should be used in both places.

BridgeSDK/BridgeAPI/SBBActivityManager.m
```diff
@@ -181,9 +180,16 @@ - (NSURLSessionTask *)getScheduledActivitiesForDaysAhead:(NSInteger)daysAhead da
...
+    // If caching, always request the maximum days ahead from the server so we have them cached
+    // and if the desiredDaysAhead is less that the max allowed, then fetch a minimum number of
+    // items per schedule.
+    NSInteger desiredDaysAhead = gSBBUseCache ? gSBBCacheDaysAhead : daysAhead;
+    NSInteger fetchDaysAhead = MIN(kMaxAdvance, desiredDaysAhead);
+    NSInteger fetchMinimumPerSchedule = (fetchDaysAhead < desiredDaysAhead) ? kMaxAdvance : 0;
+    NSDictionary *parameters = @{@"daysAhead": @(fetchDaysAhead),
+                                 @"minimumPerSchedule": @(fetchMinimumPerSchedule),
+                                 @"offset": [[NSDate date] ISO8601OffsetString]};
+    return [self.networkManager get:kSBBActivityAPI headers:headers parameters:parameters completion:^(NSURLSessionTask *task, id responseObject, NSError *error) {
```
syoung-smallwisdom: This is the code that limits the fetch values sent to the server and sets the max to 4 days. If the app is requesting more than 4 days ahead, then the code will fall back to requesting "up to 4 activities per schedule" (for those schedules that do not return any activities within the next 4 days.)

For example:

In Smart4SURE, all schedules are either 1 week and 12 weeks. So they will all return 4 activities per schedule. This is then filtered by the requested daysAhead and daysBehind before returning in the completion handler. (In that case, with a 10 day window, it will always filter to no more than 1 schedule per activity)

In mPower, where activities are scheduled 3x daily, the server will return no more than 4 days out for those schedules (or 12 activities) but will will still return a monthly activity which could be displayed in the "coming up" section of the activities view.

BridgeSDK/BridgeSDK.m
syoung-smallwisdom: In the words of the Rolling Stones, "You can't always get what you want".  :)

Alx was concerned that allowing for a longer "time in advance" could have a serious performance hit for applications that have a lot of scheduled activities. Because of this, he implemented logic as follows:

1. Get up to 4 days ahead of schedules, based on the value of the "daysAhead" parameter.
2. If the number of returned activities for a given schedule is less than "minimumPerSchedule" then return a maximum of that number of activities for each schedule.

Both of these numbers are capped at 4 and the server will return an error if you send a value greater than that.  However, in order to make his server logic work with the existing caching architecture that we have in place (which uses a simpler design of a time-banded range of days ahead and days behind), I had to fudge it.

While we could re-architect the caching to use the same daysAhead/daysBehind/minimumPerSchedule architecture, that was more effort AND I think it's confusing.  So instead I hide these implementation details.

The purpose of `SBBMaxSupportedCacheDays` is really intended to limit the "days behind" caching range since the current architecture (of grabbing the entire cache of schedules and then writing it back to Core Data rather than using a more complicated data sync algorithm) would cause performance issues if the number of records gets too big. I felt that solving that problem was a different issue that could be prioritized in JIRA with it's own ticket.

This pull request is intended to address the more immediately pressing issue of "Smart4SURE needs to be able to get 10 days ahead and the server only supports 4 days ahead or a maximum of 4 activities per schedule".

BridgeSDK/BridgeAPI/SBBActivityManager.m
```diff
@@ -181,9 +180,16 @@ - (NSURLSessionTask *)getScheduledActivitiesForDaysAhead:(NSInteger)daysAhead da
     NSMutableDictionary *headers = [NSMutableDictionary dictionary];
     [self.authManager addAuthHeaderToHeaders:headers];
     
-    // if caching, always request the maximum days ahead from the server so we have them cached
-    NSInteger fetchDaysAhead = gSBBUseCache ? kMaxAdvance : daysAhead;
-    return [self.networkManager get:kSBBActivityAPI headers:headers parameters:@{@"daysAhead": @(fetchDaysAhead), @"offset": [[NSDate date] ISO8601OffsetString]} completion:^(NSURLSessionTask *task, id responseObject, NSError *error) {
+    // If caching, always request the maximum days ahead from the server so we have them cached
+    // and if the desiredDaysAhead is less that the max allowed, then fetch a minimum number of
```
Erin-Mounts: should this say "more than," then?

BridgeSDK/BridgeAPI/SBBActivityManager.m
Erin-Mounts: makes sense

Erin-Mounts:
One last question about a code comment, otherwise looks good; merge when ready

PR #97, opened 2016-10-15
fix bug in setupWithStudy:environment:

Erin-Mounts:
Deprecated method was rewritten to call the new canonical method, but inadvertently called it with the default environment rather than the value passed in. Discovered while updating Sarcoidosis to latest libraries.

Also bump module version.


syoung-smallwisdom:
Looks good

PR #105, opened 2017-01-04
Update the method signature for signup.

syoung-smallwisdom:


Erin-Mounts:
whoops, looks good, thanks

PR #109, opened 2017-01-18
Added SBBBridgeInfoProtocol to the project.

syoung-smallwisdom:
Pulled out stuff related to setting up bridge into a protocol that can be used to setup the study.

### DIFF
--- a/BridgeSDK/BridgeAPI/SBBBridgeInfoProtocol.h
+++ b/BridgeSDK/BridgeAPI/SBBBridgeInfoProtocol.h
@@ -61,6 +61,21 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, readonly) SBBEnvironment environment;
 
+/**
+ Keychain service name.
+ */
+@property (nonatomic, readonly, copy) NSString * _Nullable keychainService;
+
+/**
+ Keychain access group name.
+ */
+@property (nonatomic, readonly, copy) NSString * _Nullable keychainAccessGroup;
+
+/**
+ App group identifier used for the suite name of NSUserDefaults (if provided).
+ */
+@property (nonatomic, readonly, copy) NSString * _Nullable appGroupIdentifier;
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/BridgeSDK/BridgeAPI/SBBBridgeInfoProtocol.h
+++ b/BridgeSDK/BridgeAPI/SBBBridgeInfoProtocol.h
@@ -1,0 +1,66 @@
+//
+//  SBBBridgeInfoProtocol.h
+//  BridgeSDK
+//
+// Copyright (c) 2017, Sage Bionetworks. All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification,
+// are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this
+// list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice,
+// this list of conditions and the following disclaimer in the documentation and/or
+// other materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder(s) nor the names of any contributors
+// may be used to endorse or promote products derived from this software without
+// specific prior written permission. No license is granted to the trademarks of
+// the copyright holders even if such marks are included in this software.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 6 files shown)


BridgeSDK/BridgeAPI/SBBBridgeInfoProtocol.h
```diff
@@ -0,0 +1,66 @@
...
+/**
+ If using BridgeSDK's built-in caching, number of days behind to cache
+ */
+@property (nonatomic, readonly) NSInteger cacheDaysBehind;
+
+/**
+ Environment to load
+ */
+@property (nonatomic, readonly) SBBEnvironment environment;
+
```
Erin-Mounts: Also need the appGroupIdentifier in this protocol (optional), for setting up the shared container and sharing NSUserDefaults and keychain when running in an app extension

BridgeSDK/BridgeAPI/SBBBridgeInfoProtocol.h
```diff
@@ -0,0 +1,66 @@
...
+/**
+ Study identifier used to setup the study with Bridge
+ */
+@property (nonatomic, readonly, copy) NSString *studyIdentifier;
+
+/**
+ Name of .pem certificate file to use for uploading to Bridge (without the .pem extension)
+ */
+@property (nonatomic, readonly, copy) NSString * _Nullable certificateName;
```
Erin-Mounts: why _Nullable?

BridgeSDK/BridgeSDK.m
```diff
@@ -40,6 +40,11 @@
 
 @implementation BridgeSDK
 
++ (void)setupWithBridgeInfo:(id <SBBBridgeInfoProtocol>)bridgeInfo {
+    gSBBPemCertificateName = bridgeInfo.certificateName;
```
Erin-Mounts: I'm actually going to be handling this a different way (copying/normalizing bridgeInfo values to a shared instance of a concrete class, then accessing the values via that shared instance throughout BridgeSDK instead of a bunch of separate globals). So no need to define/set this global--just ignore for now.

BridgeSDK/BridgeAPI/SBBBridgeInfoProtocol.h
syoung-smallwisdom: Because that is what it was defined as in `SBABridgeInfo` (for reasons of reverse compatibility to mPower)

Erin-Mounts: fair enough.

BridgeSDK/BridgeAPI/SBBBridgeInfoProtocol.h
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ Environment to load
+ */
+@property (nonatomic, readonly) SBBEnvironment environment;
+
+/**
+ Keychain service name.
+ */
+@property (nonatomic, readonly, copy) NSString * _Nullable keychainService;
```
Erin-Mounts: After noodling on it for a while, I don't think we want this keychainService and keychainAccessGroup in BridgeSDK after all. If the app developer wants to let Bridge handle storing the credentials, they can add BridgeSDK's keychain access group (org.sagebase.Bridge) to both the app and extension targets. If they want to provide an auth delegate, they just need to share the auth delegate code between the app and the extension (so they both look in the same place) and can use whatever keychain group they want, and Bridge doesn't need to know anything about it.

If at some point in the future someone wants to not use an auth delegate but still wants to be able to specify a different keychain access group for BridgeSDK to use in either or both the app and the extension for whatever reason (say they have multiple apps using BridgeSDK without an auth delegate and don't want the instances of BridgeSDK to share keychains), we should add a new pair of keys for that and not just use these; I'd rather keep the BridgeSDK keychain separate from whatever keychain the developer is using for their own purposes.

Does this make sense, or am I still overlooking something?

syoung-smallwisdom: I can take it out for now but I'm pretty sure you're gonna need it. Just please keep the property names the same as that in `SBASharedAppInfo` if you add it back in once you figure out how you are going to manage the keyboard and host app sharing a keychain.

Erin-Mounts: Well, my assumption was that they'd use something like this between themselves (via BridgeAppSDK in the app, and an extension to the SBBBridgeInfoProtocol in the keyboard that adds the same members). It just doesn't make sense for BridgeSDK itself to know about these because it wouldn't use them itself (only indirectly via the auth delegate shared between the app and the keyboard).

Erin-Mounts:
looks good

PR #111, opened 2017-01-20
Make emailForAuthManager: required too

Erin-Mounts:
what the heck, can’t think of a single good reason not to beyond "it was optional before"

### DIFF
--- a/BridgeSDK/BridgeAPI/SBBAuthManager.h
+++ b/BridgeSDK/BridgeAPI/SBBAuthManager.h
@@ -94,10 +94,22 @@ extern  NSString * _Nonnull gSBBAppStudy;
 - (void)authManager:(nullable id<SBBAuthManagerProtocol>)authManager didGetSessionToken:(nullable NSString *)sessionToken forEmail:(nullable NSString *)email andPassword:(nullable NSString *)password;
 
 /*!
+ *  This delegate method should return the email for the user account last signed up for or signed in to,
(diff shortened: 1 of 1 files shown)


BridgeSDK/BridgeAPI/SBBAuthManager.h
```diff
@@ -94,30 +94,30 @@ extern  NSString * _Nonnull gSBBAppStudy;
...
+ *  This delegate method should return the email for the user account last signed up for or signed in to,
  *  or nil if the user has never signed up or signed in on this device.
  *
- *  @note This method is now required. The password is used when encrypting sensitive user data in CoreData, and also (if emailForAuthManager: is implemented as well) for refreshing the session token automatically when 401 status codes are received from the Bridge API.
+ *  @note This method is now required, so that the SDK can handle refreshing the session token automatically when 401 status codes are received from the Bridge API.
  *
  *  @param authManager The auth manager instance making the delegate request.
  *
- *  @return The password, or nil.
+ *  @return The username, or nil.
```
syoung-smallwisdom: email rather than user name? or is this just git confusion?

Erin-Mounts: no, that's from when we changed username to email in the Bridge API ages ago and I copypasted the old method declaration and doc comments and missed this. Good catch, thanks.

PR #112, opened 2017-01-27
App extension support

Erin-Mounts:


BridgeSDK/BridgeAPI/SBBBridgeInfo.m
```diff
@@ -0,0 +1,155 @@
...
+    if (self = [super init]) {
+        _bridgeInfo = [NSMutableDictionary dictionary];
+    }
+    return self;
+}
+
+- (instancetype)initWithDictionary:(NSDictionary *)dictionary
+{
+    if (self = [super init]) {
+        _bridgeInfo = [dictionary mutableCopy];
```
syoung-smallwisdom: Why the mutable dictionary rather than just using `-setValuesForKeysFromDictionary` since these properties are all read-write?

BridgeSDK/Networking/SBBNetworkManager.m
```diff
@@ -118,7 +116,8 @@ @interface SBBNetworkManager () <NSURLSessionDataDelegate, NSURLSessionDownloadD
 @property (nonatomic, strong) NSURLSession * mainSession; //For data tasks
 @property (nonatomic, strong) NSURLSession * backgroundSession; //For upload/download tasks
 
-@property (nonatomic, copy) void (^backgroundCompletionHandler)(void);
+//@property (nonatomic, copy) void (^backgroundCompletionHandler)(void);
```
syoung-smallwisdom: commented out line?

BridgeSDK/BridgeAPI/SBBBridgeInfo.h
```diff
@@ -0,0 +1,39 @@
...
+//	SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+#import <BridgeSDK/SBBBridgeInfoProtocol.h>
+
+@interface SBBBridgeInfo : NSObject <SBBBridgeInfoProtocol>
+
++ (instancetype)shared;
+
+- (void)setFromBridgeInfo:(id<SBBBridgeInfoProtocol>)info;
```
Erin-Mounts: yep--was avoiding that because of old obj-c naming conventions related to memory management, which even after ARC still used to give static analyzer warnings--but apparently not anymore, so fixing

BridgeSDK/BridgeAPI/SBBBridgeInfo.m
Erin-Mounts: mainly because _bridgeInfo isn't set at this point so that wouldn't do anything, but that's easily fixed if I call [self init] instead of [super init] on the line above.

BridgeSDK/Networking/SBBNetworkManager.m
Erin-Mounts: removing, thanks

PR #119, opened 2017-02-07
Don’t use standardUserDefaults by default

Erin-Mounts:
unless there’s a flag to that effect in BridgeInfo in the setup call

### DIFF
--- a/BridgeSDK/BridgeAPI/SBBBridgeInfo.m
+++ b/BridgeSDK/BridgeAPI/SBBBridgeInfo.m
@@ -88,6 +88,8 @@ const NSInteger SBBMaxSupportedCacheDays = 30;
     self.environment = info.environment;
     self.certificateName = info.certificateName;
     self.appGroupIdentifier = info.appGroupIdentifier;
+    self.usesStandardUserDefaults = info.usesStandardUserDefaults;
+    self.userDefaultsSuiteName = info.userDefaultsSuiteName;
 }
 
 - (NSString *)studyIdentifier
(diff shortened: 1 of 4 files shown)


BridgeSDK/BridgeAPI/SBBBridgeInfo.m
```diff
@@ -152,4 +152,24 @@ - (void)setAppGroupIdentifier:(NSString *)appGroupIdentifier
...
+
+- (NSString *)userDefaultsSuite
+{
+    return _bridgeInfo[NSStringFromSelector(@selector(userDefaultsSuite))];
+}
+
+- (void)setUserDefaultsSuite:(NSString *)userDefaultsSuite
+{
+    _bridgeInfo[NSStringFromSelector(@selector(userDefaultsSuite))] = [userDefaultsSuite copy];
+}
```
syoung-smallwisdom: Don't these also need to be added to the `copyFromBridgeInfo:` method?

BridgeSDK/BridgeAPI/SBBBridgeInfoProtocol.h
```diff
@@ -66,9 +66,25 @@ NS_ASSUME_NONNULL_BEGIN
...
 
 /**
- App group identifier used for the suite name of NSUserDefaults, and for the name of the shared
- container, which is used both to configure the background session and as the place to store
- temporary copies of files being uploaded to Bridge (if provided).
+ Tells the Bridge libraries to use the standard user defaults suite.
+ 
+ @note This flag is intended only for backward compatibility when upgrading apps built with older versions of Bridge libraries that used the standard user defaults suite. It will be ignored in any case if either userDefaultsSuite or appGroupIdentifier are set.
+ */
+@property (nonatomic, readonly) BOOL useStandardUserDefaults;
```
syoung-smallwisdom: Does this mean that all our existing apps would have to set this flag to true?

Erin-Mounts: yes, if we want them to keep what they had stored in NSUserDefaults. I just think that by default, going forward, we ought not to be using standard defaults, so we can manage ours more easily without interfering with settings we don't know or care about that the app developer uses.

BridgeSDK/BridgeAPI/SBBBridgeInfo.m
Erin-Mounts: oops, yes

BridgeSDK/BridgeAPI/SBBBridgeInfoProtocol.h
syoung-smallwisdom: Makes sense, just wanted to make sure I was reading that correctly.

PR #120, opened 2017-02-08
Fix openssl build

Erin-Mounts:
- adding pre-built libOpenSSL.a and OpenSSL/ header files to the repo, and then linking with the pre-built lib instead of the one in the built products folder
- adding a libOpenSSL build target to BridgeSDK, to manually rebuild and recopy the lib and headers if needed


Erin-Mounts:
closing because I just noticed the bitcode flag isn't being set when the library is built

syoung-smallwisdom:
Also, I don't see where you removed CMSSupport?

PR #125, opened 2017-03-01
Remove docs from git

zaro0508:
Documentation is a built artifact and should not be committed to
the repo.  It should be generated from CI build and published to
a remote location for others to consume.


Erin-Mounts:
Looks good.

PR #130, opened 2017-03-02
Share documentation scheme

zaro0508:
Set xcode to share the documentation scheme so CI build slaves have
access to it.


Erin-Mounts:
looks good

PR #131, opened 2017-03-03
Update participant sharing...

Erin-Mounts:
...in cached StudyParticipant when it gets set while recording consent to Bridge.


syoung-smallwisdom:
Looks good

PR #135, opened 2017-03-20
Change the documentation deployment location

zaro0508:
SDK are hosted on https://developer.sagebridge.org so deployments so we
should deploy the docs there.  Also only deploy docs when a tag is made.


zaro0508:
I'm assuming that tags are created when application releases happen therefore this is setup to deploy docs when tags are created.

PR #138, opened 2017-04-13
Get credentials from env vars if not in plist

Erin-Mounts:
so CI can run the integration tests without checking the credentials plist into Github anywhere

### DIFF
--- a/BridgeSDKTests/SBBBridgeAPIIntegrationTestCase.m
+++ b/BridgeSDKTests/SBBBridgeAPIIntegrationTestCase.m
@@ -49,23 +49,40 @@ NSString * const kUserSessionInfoIdKey = @"id";
         
         // sensitive credentials are stored in a plist file that lives *outside* of the local git repo
         NSString *credentialsPlist = [[NSBundle bundleForClass:[self class]] pathForResource:@"BridgeAdminCredentials" ofType:@"plist"];
-        NSDictionary *credentials = [[NSDictionary alloc] initWithContentsOfFile:credentialsPlist][@"studies"];
-        NSDictionary *studyCredentials = credentials[[SBBBridgeInfo shared].studyIdentifier];
+        NSString *credentialsEmail = nil;
+        NSString *credentialsPassword = nil;
+        if (credentialsPlist) {
+            NSDictionary *credentials = [[NSDictionary alloc] initWithContentsOfFile:credentialsPlist][@"studies"];
+            NSDictionary *studyCredentials = credentials[[SBBBridgeInfo shared].studyIdentifier];
+            credentialsEmail = studyCredentials[@"email"];
+            credentialsPassword = studyCredentials[@"password"];
+        }
         
-        [gAdminAuthManager signInWithEmail:studyCredentials[@"email"] password:studyCredentials[@"password"] completion:^(NSURLSessionTask *task, id responseObject, NSError *error) {
-            if (error) {
-                NSLog(@"Error logging in to admin account:\n%@\nResponse: %@", error, responseObject);
-            } else {
-                NSLog(@"Logged in to admin account:\n%@", responseObject);
-            }
-            [expectAdminSignin fulfill];
-        }];
+        // ...if not found, check env vars
+        if (!credentialsEmail.length || !credentialsPassword.length) {
+            NSDictionary *environment = NSProcessInfo.processInfo.environment;
+            credentialsEmail = credentialsEmail.length ? credentialsEmail : environment[@"SAGE_ADMIN_EMAIL"];
+            credentialsPassword = credentialsPassword.length ? credentialsPassword : environment[@"SAGE_ADMIN_PASSWORD"];
+        }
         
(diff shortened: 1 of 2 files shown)


BridgeSDKTests/SBBBridgeAPIIntegrationTestCase.m
```diff
@@ -49,23 +49,40 @@ - (void)setUp {
...
-        [gAdminAuthManager signInWithEmail:studyCredentials[@"email"] password:studyCredentials[@"password"] completion:^(NSURLSessionTask *task, id responseObject, NSError *error) {
-            if (error) {
-                NSLog(@"Error logging in to admin account:\n%@\nResponse: %@", error, responseObject);
-            } else {
-                NSLog(@"Logged in to admin account:\n%@", responseObject);
-            }
-            [expectAdminSignin fulfill];
-        }];
+        // ...if not found, check env vars
+        if (!credentialsEmail.length || !credentialsPassword.length) {
```
tcisagebio: It looks like the idea is to override email and password if there's an env for it.  However it looks like you still need to provide a plist file?

Also should these be checked separately?  What if SAGE_ADMIN_PASSWORD is set but not SAGE_ADMIN_EMAIL?

tcisagebio: shouldn't the env vars override the plist setting?  i'm thinking in the case where both are provided.

Erin-Mounts: can't think of a reason to prefer one over the other, really, and in general you're either going to be running in the GUI on a local machine, in which case you have to use the plist values or nothing, or you're going to be running the tests on a travis/fastlane machine and won't have the plist available except as a stub.

zaro0508: sorry, i shouldn't have used the bot user to review.  Not a big deal but a typical use case would be to use a plist file as the default then make quick changes by overriding the values from the command line. 

Also on line 54 it looks like *credentials and *studyCredentials is only created if you provide a plist file?  Would this still work if you only provide the environment vars?

Erin-Mounts: would you really be overriding valid credentials from the plist file with different valid credentials from the command line/environment?

yes those variables are only set if you provide a plist file, but they're only used to get the values out of the plist and not needed otherwise.

Erin-Mounts: actually can't see a problem with just overriding from environment if those exist, so will make that change

Erin-Mounts: ok, changed it to look in env vars first, and then if either email or password is missing from there, to get both from the plist (doesn't really make sense to ever get email from one place and password from the other, since credentials are a matched set).

BridgeSDK.xcodeproj/xcshareddata/xcschemes/BridgeSDK.xcscheme
```diff
@@ -58,6 +58,18 @@
...
       </MacroExpansion>
+      <EnvironmentVariables>
+         <EnvironmentVariable
+            key = "SAGE_ADMIN_EMAIL"
+            value = "bridgeit@sagebase.org"
+            isEnabled = "YES">
+         </EnvironmentVariable>
+         <EnvironmentVariable
+            key = "SAGE_ADMIN_PASSWORD"
+            value = "apB988xEUv&lt;"
```
tcisagebio: committing a valid password?

Erin-Mounts: dangit, I reverted that before making the original PR commit. How did it come back? Well, fixed (again)

BridgeSDKTests/SBBBridgeAPIIntegrationTestCase.m
tcisagebio: I would prefer env vars to have precedence over file.  I think that would make testing easier since I can just pass in a var when i run a command instead of having to change a file.  for example:

SAGE_ADMIN_EMAIL="foo@sage.com" SAGE_ADMIN_PASSWORD="bar" fastlane mytest

Erin-Mounts: Yes, that's what this does. It only even looks at the plist if one or both credentials is missing from the env vars.

BridgeSDK.xcodeproj/xcshareddata/xcschemes/BridgeSDK.xcscheme
zaro0508: password has been committed to a public repo.  maybe time to change it?

Erin-Mounts: yep. dang again.

PR #151, opened 2017-05-27
Cache wins for writable fields unless nil.

Erin-Mounts:
Fixes IA-120


dephillipsmichael:
This looks like it does solve the problem of cache wins for client data, but I'm still not seeing the ScheduledActivityManager return the correct clientData.  There must be something else going on.  I'll look into it deeper and see if I can get more information.

dephillipsmichael:
Ok, so it appears that when "reconcileWithDictionaryRepresentation" is called, the clientData is empty, I'm going to continue investigating.

PR #156, opened 2017-06-17
Participant profile support

Erin-Mounts:


BridgeSDK/BridgeAPI/SBBParticipantManager.m
```diff
@@ -40,6 +40,21 @@ + (instancetype)defaultComponent
...
++ (NSArray<NSString *> *)dataSharingScopeStrings
+{
+    static NSArray<NSString *> *scopeStrings = nil;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        scopeStrings =  @[
+                          kSBBParticipantDataSharingScopeStrings[0],
+                          kSBBParticipantDataSharingScopeStrings[1],
+                          kSBBParticipantDataSharingScopeStrings[2]
+                          ];
```
syoung-smallwisdom: This feels like a maintenance crash waiting to happen. Could you please build using `sizeof`?

BridgeSDK/BridgeSDK.m
```diff
@@ -163,6 +165,27 @@ + (BOOL)restoreBackgroundSession:(NSString *)identifier completionHandler:(void
...
+            unsigned int numProperties;
+            objc_property_t *properties = class_copyPropertyList(SBBStudyParticipantCustomAttributes.class, &numProperties);
+            free(properties);
+
+            if (!info.studyParticipant.attributes && numProperties > 0) {
+                info.studyParticipant.attributes = [[SBBStudyParticipantCustomAttributes alloc] init];
+            }
+        }
+        [delegate authManager:nil didReceiveUserSessionInfo:info];
+    }
```
syoung-smallwisdom: There is a call made on logout or during onboarding to flush the temp state info that is stored if the user cancels (`resetStoredUserData`). That method should call some method to flush this cache as well.

BridgeSDK/BridgeAPI/SBBAuthManager.m
```diff
@@ -296,14 +297,19 @@ - (NSString *)sessionToken
...
+    NSString *studyParticipantType = SBBStudyParticipant.entityName;
+    SBBStudyParticipant *studyParticipant = (SBBStudyParticipant *)[SBBComponent(SBBCacheManager) cachedSingletonObjectOfType:studyParticipantType createIfMissing:NO];
+    NSMutableDictionary *json = [NSMutableDictionary dictionary];
+    if (studyParticipant) {
+        json = [[SBBComponent(SBBObjectManager) bridgeJSONFromObject:studyParticipant] mutableCopy];
+    }
+    
+    // now overwrite with entries from signUp object
+    NSDictionary *signUpJson = [signUp dictionaryRepresentation];
+    [json addEntriesFromDictionary:signUpJson];
```
Erin-Mounts: Participant info is cleared from the cache upon signOut, and the auth delegate method is called with nil for the userSessionInfo.

BridgeSDK/BridgeAPI/SBBParticipantManager.m
Erin-Mounts: yes--what was I thinking... thanks

BridgeSDK/BridgeAPI/SBBAuthManager.m
syoung-smallwisdom: Yes, but what about the interim state of onboarding? If the user cancels onboarding then their locally cached profile info needs to be cleared, right?

BridgeSDK/BridgeSDK.m
Erin-Mounts: yes--just trying to figure out the right way to expose that to BridgeAppSDK, because calling it in just about any other context would break a lot of stuff.

PR #160, opened 2017-06-23
V4 activities

Erin-Mounts:
fixes https://sagebionetworks.jira.com/browse/IA-93


Erin-Mounts:
also updates Survey model with rules available on survey elements, etc.

PR #161, opened 2017-06-23
SurveyElement.rules -> .afterRules

Erin-Mounts:


syoung-smallwisdom:
Um... this is going to break BridgeAppSDK, isn't it?

Erin-Mounts:
I’m changing BridgeAppSDK to use the v4 activities APIs as we “speak”, so I could just change that property name while I’m at it—or is there more to it than that?

Erin-Mounts:
also--adding rules to SurveyElement already breaks BridgeAppSDK; changing it to afterRules would keep it from breaking: 
"/Users/erin/Development/Sage/BridgeAppSDK/BridgeAppSDK/SBASurveyItem+Bridge.swift:249:16: Property 'rules' with type '[SBASurveyRuleItem]?' cannot override a property with type '[Any]!'"

Erin-Mounts:
to clarify--I'm not changing the rules object down in the constraints, just the new one I added at the SurveyElement (question or information) level that BridgeAppSDK wasn't aware of yet anyway.

PR #163, opened 2017-06-27
IA-286 and fix filtering

Erin-Mounts:


BridgeSDK/BridgeAPI/BridgeAPITypes/SBBDateTimeRangeResourceList.m
```diff
@@ -55,17 +55,29 @@ - (void)reconcileWithDictionaryRepresentation:(NSDictionary *)dictionary objectM
...
-                              comparisonKey, self.startTime,
-                              comparisonKey, self.endTime];
+    NSString *scheduledOnKey = NSStringFromSelector(@selector(scheduledOn));
+    NSString *expiresOnKey = NSStringFromSelector(@selector(expiresOn));
+    
+    NSPredicate *predicate = [NSPredicate predicateWithFormat:@"(%K < %@ && %K != nil && %K < %@) OR %K >= %@",
+                              scheduledOnKey, self.startTime,
+                              expiresOnKey,
+                              expiresOnKey, self.startTime,
+                              scheduledOnKey, self.endTime];
```
syoung-smallwisdom: You seem to be missing the finished on part of this. You also need to filter on `finishedOn`, don't you? 

````
NSPredicate *finishedOutsideRange = [NSPredicate predicateWithFormat:@"%K != nil AND (%K < %@ OR %K >= %@)", finishedOnKey, finishedOnKey, self.startTime, finishedOnKey, self.endTime ];
NSPredicate *scheduledOutsideRange = [NSPredicate predicateWithFormat:@"(%K < %@ && %K != nil && %K < %@) OR %K >= %@",
                              scheduledOnKey, self.startTime,
                              expiresOnKey,
                              expiresOnKey, self.startTime,
                              scheduledOnKey, self.endTime];
NSPredicate *predicate = [NSCompoundPredicate orPredicateWithSubpredicates: @[scheduledOutsideRange, finishedOutsideRange]];
````

You might want to take a look at the predicates that I created in an extension on `SBBScheduledActivity`. Specifically, this one:

````
    public static func scheduledPredicate(on date:Date) -> NSPredicate {
        let startOfDay = date.startOfDay()
        let startOfNextDay = startOfDay.addingNumberOfDays(1)
        
        // Scheduled for this date or prior
        let scheduledKey = #keyPath(scheduledOn)
        let scheduledThisDayOrBefore = NSPredicate(format: "%K == nil OR %K < %@", scheduledKey, scheduledKey, startOfNextDay as CVarArg)
        let unfinished = unfinishedPredicate()
        let finishedOnThisDay = finishedPredicate(on: date)
        
        let expiredKey = #keyPath(expiresOn)
        let expiredOnThisDay = NSPredicate(format: "%K == nil OR (%K >= %@ AND %K < %@)", expiredKey, expiredKey, startOfDay as CVarArg, expiredKey, startOfNextDay as CVarArg)
        let expiredOnOrAfterThisDay = NSPredicate(format: "%K == nil OR %K > %@", expiredKey, expiredKey, startOfDay as CVarArg)
        
        switch(startOfDay.compare(Date().startOfDay())) {
            
        case .orderedAscending:
            // build a filter for a day in the past that includes expired on that day OR completed on that day
            let expired = NSCompoundPredicate(andPredicateWithSubpredicates: [unfinished, expiredOnThisDay])
            let finishedOrExpired = NSCompoundPredicate(orPredicateWithSubpredicates: [finishedOnThisDay, expired])
            return NSCompoundPredicate(andPredicateWithSubpredicates: [scheduledThisDayOrBefore, finishedOrExpired])
        
        case .orderedSame:
            // build a filter for today that includes activites completed today, expiring today or later and scheduled to 
            // include today
            let unfinishedOrFinishedToday = NSCompoundPredicate(orPredicateWithSubpredicates: [unfinished, finishedOnThisDay])
            return NSCompoundPredicate(andPredicateWithSubpredicates: [scheduledThisDayOrBefore, unfinishedOrFinishedToday, expiredOnOrAfterThisDay])
        
        case .orderedDescending:
            // For the future, we only want unfinished schedules
            return NSCompoundPredicate(andPredicateWithSubpredicates: [scheduledThisDayOrBefore, unfinished, expiredOnOrAfterThisDay])
        }
    }
````

File: BridgeSDK/BridgeAPI/SBBActivityManager.m [hunk 2]
```diff
@@ -116,12 +116,18 @@ - (NSArray *)filterTasks:(NSArray *)tasks forDaysAhead:(NSInteger)daysAhead andD
 
 - (NSArray *)filterTasks:(NSArray *)tasks scheduledFrom:(NSDate *)startDate to:(NSDate *)endDate
 {
+    // It's in the date range [startDate, endDate) if:
+    //  - it never expires or expires on or after startDate AND
+    //  - it hasn't been marked finished or was marked finished on or after startDate AND
+    //  - it was scheduled before endDate
```
syoung-smallwisdom: Okay, now I'm confused?  Why are you checking finished on here but not above? Also, you have an assumption here that the date range isn't fully within the past. For instance a UI where today is filtered out (for whatever reason) would fail this test. For example, if the task was finished today but the requested range only includes yesterday.

BridgeSDK/BridgeAPI/BridgeAPITypes/SBBDateTimeRangeResourceList.m
Erin-Mounts: whoops, yes, missed that. I was up waaay past my bedtime. :P

BridgeSDK/BridgeAPI/SBBActivityManager.m
Erin-Mounts: This will return all (and only) scheduled activities that existed in a scheduled, but not finished or expired, state at any point during the requested span, following the logic of the server API. BridgeSDK doesn't know the app's intentions with the results (e.g. whether it's going to use them to show activities that are available and still need to be done or to show past activity results), so if for example the app wants all activities that were finished between a month ago and yesterday, it would need to get everything that existed in that time span and then filter out or ignore the ones that got finished after the end of yesterday, and a UI that wants to show available activities to do would need to filter out or ignore the ones already marked finished.

BridgeSDK/BridgeAPI/BridgeAPITypes/SBBDateTimeRangeResourceList.m
syoung-smallwisdom: It happens. ;)

BridgeSDK/BridgeAPI/SBBActivityManager.m
syoung-smallwisdom: got it. Thanks.

BridgeSDK/BridgeAPI/BridgeAPITypes/SBBDateTimeRangeResourceList.m
Erin-Mounts: ok, as it turns out, this logic is actually correct for matching what the server returns for a given date range (and it needs to do that precisely here, for this to work).

syoung-smallwisdom: Ok...

PR #167, opened 2017-08-25
more_mime_types

dephillipsmichael:
Added more mime types that can be automatically converted.
Is there anyway to inject these?  If not, I hope we can add these.


dephillipsmichael:
I don't see how adding a few branches to an existing conditional could break the build?  Looking into why CI failed....

dephillipsmichael:
I see the test that's failing, but I can't run tests locally for more info because of Xcode can't find mogenerator?

Erin-Mounts:
so, pull mogenerator from my fork here: https://github.com/Erin-Mounts/mogenerator.git then open the project file and build it; that should install it where the build script is looking for it.

dephillipsmichael:
Closing until we get back to Resilience

PR #175, opened 2017-12-20
update to support new devices

Erin-Mounts:
since the last time this was updated


Erin-Mounts:
also removed an integration test that now always fails because the endpoint requires a verified email address.

PR #181, opened 2018-02-14
Fix for research suite

syoung-smallwisdom:
I added string enums for the keywords used by bridge.

Additionally, I changed the "title" property from non-nil to nullable. This is to support `RSDUIStep.title` which can be nil. Since these are never set *to* the server and are only pulled *from* the server, this should not be a problem for our services.


BridgeSDK/SBBDefines.h
```diff
@@ -0,0 +1,38 @@
+//
+//  SBBDefines.h
+//  BridgeSDK
+//
+//  Copyright © 2016 Sage Bionetworks. All rights reserved.
```
Erin-Mounts: 2018

Erin-Mounts:
Oh, also should bump the build number with agvtool.

syoung-smallwisdom:
@Erin-Mounts ping.  I think I have all the changes. Thanks!

PR #182, opened 2018-02-16
Add schedule plan guid and placeholders for "nice-to-have" properties

syoung-smallwisdom:
I added the schedule plan guid, but did not try to put in logic for a SBBSchedulePlan object. I figure this is something that we will want to consider for future builds but would not prioritize for round 1.

Additionally, I added properties to SBBActivity references for additional information support in case we ever add these kinds of properties to the references (rather than merging from a custom clientData blob).


BridgeSDK/BridgeAPI/BridgeAPITypes/SBBDataModel.xcdatamodeld/SBBDataModel.xcdatamodel/contents
```diff
@@ -164,6 +171,9 @@
                 <entry key="notInPONSODictionary" value="1"/>
             </userInfo>
         </relationship>
+        <userInfo>
```
Erin-Mounts: I don't think this flag does anything here (at the entity level)...?

PR #185, opened 2018-03-20
Schemas round2

syoung-smallwisdom:
Add a `copy(with:)` method to the `SBBActivity` to support `RSDTaskInfo` protocol.

### DIFF
--- a/BridgeSDK/BridgeAPI/BridgeAPITypes/SBBActivity.m
+++ b/BridgeSDK/BridgeAPI/BridgeAPITypes/SBBActivity.m
@@ -33,6 +33,9 @@
 
 #pragma mark Abstract method overrides
 
-// Custom logic goes here.
+- (instancetype)copyWithIdentifier: (NSString *)identifier {
+    SBBActivity *copy = [self copy];
+    return copy;
+}
 
 @end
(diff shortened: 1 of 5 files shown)


BridgeSDK/BridgeAPI/BridgeAPITypes/SBBActivity.m
```diff
@@ -33,6 +33,9 @@ @implementation SBBActivity
 
 #pragma mark Abstract method overrides
 
-// Custom logic goes here.
+- (instancetype)copyWithIdentifier: (NSString *)identifier {
+    SBBActivity *copy = [self copy];
+    return copy;
```
Erin-Mounts: ...should it be updating the guid on the copy here?

syoung-smallwisdom: No. I considered that but since this is a wrapper implementation of `RSDTaskInfo`, it's not really appropriate. That said... just realized that I am not sure how `identifier` is mapped... so... let me get back to you on that. ;)

syoung-smallwisdom: @Erin-Mounts 
Updated to copy the identifier to whichever reference pointer is non-nil. Thanks!

PR #191, opened 2018-05-11
Add objc methods

syoung-smallwisdom:
These are used for building predicates that require KVO. Since these are obj-c objects, Swift is less confused when it just calls the method defined in obj-c. Seemed to make more sense to include them in BridgeSDK directly.


BridgeSDK/BridgeAPI/BridgeAPITypes/SBBScheduledActivity.m
```diff
@@ -115,4 +115,39 @@ - (void)reconcileWithDictionaryRepresentation:(NSDictionary *)dictionary objectM
...
+{
+    return self.activity.task.identifier ?: self.activity.survey.identifier ?: self.activity.compoundActivity.identifier;
+}
+
+- (NSString *) scheduleIdentifier
+{
+    NSRange range = [self.guid rangeOfString:@":"];
+    if (range.location != NSNotFound)
+    {
+        return [self.guid substringToIndex:range.location];
```
Erin-Mounts: This is actually the same as self.activity.guid, I'm like 97% sure...

There's also a self.schedulePlanGuid member now, that has the guid of the schedule plan that generated this scheduled activity, which might be what you're actually looking for though?

syoung-smallwisdom: No, the `schedulePlanGuid` is for a different object. 

There is a one-to-many for the schedulePlan to scheduledActivity, and a one-to-one for the scheduledActivity to Activity. Might be that the self.activity.guid is the same thing, but... if it ain't broke... ;)

PR #195, opened 2018-05-23
Create test setup

syoung-smallwisdom:
I added a target framework to the BridgeSDK framework that can bypass authentication and allow objects to be created without being authenticated.

I basically got it working enough so that I can replace the SBBActivityManager with my own version that uses in-memory schedules that can be updated dynamically to mimic the server.

More PR's to come for the other projects. (I'm still debugging the UI.)


BridgeSDKTests/MockURLSession.m
```diff
@@ -84,14 +84,14 @@ - (void)resume
 {
     [_session.delegateQueue addOperationWithBlock:^{
         MockHTTPURLResponse *response;
-        __unused NSData *data = [_session dataAndResponse:&response forRequest:_request];
-        _mockResponse = response;
+        __unused NSData *data = [self->_session dataAndResponse:&response forRequest:self->_request];
```
Erin-Mounts: Was this set of changes to fix warnings or a crash? Wouldn't capturing self and dereferencing in the block, instead of capturing the individual _session, _request, or _mockResponse as they existed at the time of the call, lead to unexpected behavior if the values happen to get changed between when the block was added to the delegate queue and when it got executed? Not that that would ever happen in typical use, just making sure there's a method behind this "madness"...

syoung-smallwisdom: Fix a warning. When you are using them in blocks, should use self to explicitly retain rather than implicitly retain (because you are referencing an ivar).

PR #204, opened 2018-08-15
make travis build docs on tags

zaro0508:
Travis was not building the docs when a tag is pushed therefore
the build would fail with 'No such file or directory @ dir_chdir -
Documentation'.  This update will build the docs so the deploy
will work.

### DIFF
--- a/travis/build.sh
+++ b/travis/build.sh
@@ -2,12 +2,12 @@
 set -ex
 # setup config file for integration tests
 cp BridgeAdminCredentials.plist ../BridgeAdminCredentials.plist
-if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then     # on pull requests
-    bundle exec fastlane doc scheme:"BridgeSDK"
+# build docs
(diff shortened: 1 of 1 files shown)


travis/build.sh
```diff
@@ -2,12 +2,12 @@
 set -ex
 # setup config file for integration tests
 cp BridgeAdminCredentials.plist ../BridgeAdminCredentials.plist
-if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then     # on pull requests
-    bundle exec fastlane doc scheme:"BridgeSDK"
+# build docs
+bundle exec fastlane doc scheme:"BridgeSDK"
```
Erin-Mounts: Is this expecting the BridgeSDK scheme to build docs? (because it doesn't as far as I can see) or is it telling fastlane to build the doc scheme (which it somehow knows is "Documentation") in the BridgeSDK Xcode project?

zaro0508: yes, this is the fastlane command to build the doc lane which is defined in fastlane/Fastfile

PR #209, opened 2018-08-15
Fix reauth check

Erin-Mounts:


BridgeSDK/BridgeAPI/SBBBridgeNetworkManager.m
```diff
@@ -116,7 +116,7 @@ - (NSURLSessionDownloadTask *)downloadFileFromURLString:(NSString *)urlString
 
 - (void)handleHTTPError:(NSError *)error task:(NSURLSessionTask *)task response:(id)responseObject retryObject:(SBBNetworkRetryObject *)retryObject
 {
-    if (retryObject && retryObject.retryBlock && error.code == SBBErrorCodeServerNotAuthenticated && [_authManager canAuthenticate])
+    if (retryObject && retryObject.retryBlock && error.code == SBBErrorCodeServerNotAuthenticated && ([_authManager canAuthenticate] || [_authManager isAuthenticated]))
```
syoung-smallwisdom: How can the auth manager reauthenticate if it doesn't have a `reauthToken`?

syoung-smallwisdom: And if it can re-authenticate as long as it has a session token (even though the session token is later cleared), then shouldn't that be part of the `canAuthenticate` method?

PR #212, opened 2018-08-22
Change completion block to nonnill for fetch requests

syoung-smallwisdom:
The `getReport` methods should have a non-nil completion.


BridgeSDK/BridgeAPI/SBBParticipantManager.h
```diff
@@ -203,7 +203,7 @@ typedef void (^SBBParticipantManagerCompletionBlock)(_Nullable id responseObject
  
  @return An NSURLSessionTask object so you can cancel or suspend/resume the request.
  */
-- (nullable NSURLSessionTask *)getReport:(nonnull NSString *)identifier fromTimestamp:(nonnull NSDate *)fromTimestamp toTimestamp:(nonnull NSDate *)toTimestamp completion:(nullable SBBParticipantManagerGetReportCompletionBlock)completion;
+- (nullable NSURLSessionTask *)getReport:(nonnull NSString *)identifier fromTimestamp:(nonnull NSDate *)fromTimestamp toTimestamp:(nonnull NSDate *)toTimestamp completion:(nonnull SBBParticipantManagerGetReportCompletionBlock)completion;
```
Erin-Mounts: I mean, I made these nullable since you could fire off a blind request to get them from bridge with the intention that they'll be in the cache later when you need to see the latest one. But I suppose you can always just use an empty completion handler if you want to do that. _shrug_

syoung-smallwisdom: It matched the pattern of other fetch methods within the framework. :)

PR #213, opened 2018-08-25
Clean keychain on reinstall

Erin-Mounts:
Weird iOS trivia I just discovered: If you kill the app before deleting it, the keychain and CoreData backing store both go away as expected and are empty upon reinstall. If you leave the app running and delete it from the phone, the keychain hangs around through reinstall, while CoreData (as expected) does not, and this causes problems for BridgeSDK.


BridgeSDK/BridgeAPI/SBBAuthManager.m
```diff
@@ -270,10 +273,26 @@ + (instancetype)authManagerWithBaseURL:(NSString *)baseURL
     return authManager;
 }
 
+- (void)clearOldKeychainOnReinstall
+{
+    NSUserDefaults *defaults = BridgeSDK.sharedUserDefaults;
```
syoung-smallwisdom: This is a little confusing. You mention Core data as the cause of the issue but you clear it using NSUserDefaults. Can you please update the comment to reflect that? Thanks!

PR #221, opened 2018-12-13
V2 survey archive

syoung-smallwisdom:
It has come up with both Northwestern and Penn State that they would prefer to use the v2_generic format. This is part 1 in supporting that. I set it up with v2_generic as the default and will add a configuration variable to use the v1_legacy for our existing applications.


BridgeSDK/BridgeAPI/SBBDataArchive.h
```diff
@@ -57,6 +57,27 @@ NS_ASSUME_NONNULL_BEGIN
 - (instancetype)initWithReference:(NSString *)reference
   jsonValidationMapping:(nullable NSDictionary <NSString *, NSPredicate *> *)jsonValidationMapping NS_DESIGNATED_INITIALIZER;
 
+/**
+ The name of the `v2_generic` data file. By default, `dataFilename = "answers.json`.
```
Erin-Mounts: looks like there's a stray " at the start of the default filename...?

PR #222, opened 2018-12-19
Fix Fatal Exception: [NSFileManager copyItemAtURL:toURL:options:error:]: destination URL is nil

pmkowal:
Hi, I found that `tempFileForFileURL:` method from `SBBUploadManager` raises an exception when `toURL` parameter is nil when `copyItemAtURL:toURL:error:` method is called on `FileManager`.

Please consider the following solution as it should prevent from crash and report an appropriate error.


Erin-Mounts:
Hi Pawel, apologies for taking so long to get back to you--I was on vacation over the holidays.

I'm hesitating to just accept your change as this is an internal method with a fairly constrained pathway to get to it, and I find it much easier to debug a crash like this than to have an unexplained failure to upload a file and try to track it back to an obscure error message buried somewhere in a console log.

Can you give me more details on how you encountered this crash? Under what circumstances can that parameter be nil? I might have overlooked a valid reason to handle it this way.

pmkowal:
Hi @Erin-Mounts, thank you for your time.
I see your point and I agree that the pull request just fixes the symptoms, not the root of an issue.
Answering your questions, I received information about the crash from the crash-report platform and the stack trace shows that it occurs when `uploadFileToBridge:completion:` method is called from `retryOrphanedUploadFromScratch:` or `BirdgeSDK.setupWithBridgeInfo:` method. Unfortunately I can't reproduce the issue and I don't have more details.
I think we can close the pull request for now and if I'm able to find the cause of the issue I will get back to you with a new one.

PR #223, opened 2019-02-15
Fix crash caused by race condition in context init

syoung-smallwisdom:
Because the method for initializing the `cacheIOContext` first checks to see if the context is nil and returns that context if not, need to set the ivar *after* the persistent store is set up. Otherwise, there is a race condition where the context can be returned before the persistent store model has been attached.

### DIFF
--- a/BridgeSDK/Internal/SBBCacheManager.m
+++ b/BridgeSDK/Internal/SBBCacheManager.m
@@ -675,19 +675,21 @@ void removeCoreDataQueueForPersistentStoreName(NSString *name)
 {
     if (!_cacheIOContext) {
         [self dispatchSyncToCacheManagerCoreDataQueue:^{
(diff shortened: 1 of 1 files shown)


BridgeSDK/Internal/SBBCacheManager.m
```diff
@@ -675,19 +675,21 @@ - (NSManagedObjectContext *)cacheIOContext
 {
     if (!_cacheIOContext) {
         [self dispatchSyncToCacheManagerCoreDataQueue:^{
```
syoung-smallwisdom: Note: if we ever have a project where the design includes reseting the database frequently, then the queue should be moved to *always* being called to check for `nil` since there's a possible race condition in that code as well. Given our current login/withdrawal app design, this isn't really an issue.

PR #227, opened 2019-05-07
Add missing header files to umbrella header

syoung-smallwisdom:


BridgeSDK/BridgeSDK.h
```diff
@@ -56,9 +56,11 @@ extern const unsigned char BridgeSDKVersionString[];
 #import <BridgeSDK/SBBNotificationManager.h>
 #import <BridgeSDK/SBBOAuthManager.h>
 #import <BridgeSDK/SBBObjectManager.h>
+#import <BridgeSDK/SBBScheduleManager.h>
```
Erin-Mounts: I left these two headers out on purpose, since they're extremely deprecated and no one should be using them anymore--but if they need to be included for some reason, okay I suppose

syoung-smallwisdom: I get warnings about them being public but not included

Erin-Mounts: good enough reason I guess

PR #231, opened 2019-05-14
Change the method signature for returning a cached report 

syoung-smallwisdom:
...to return an empty report if there isn't an error, but there also isn't a valid report.

The other alternative is to throw an error that indicates that there isn't any cached data, but I don't know that that is an improvement. The method this is based on returns an array if it succeeds, but that array is empty if it needs to trigger a call to the server.

See https://github.com/Sage-Bionetworks/BridgeApp-Apple-SDK/pull/131


syoung-smallwisdom:
Note: I looked into travis and it looks like its already set up to only run the unit tests and not the integration tests on the PR *but* it still tries to *build* the integration tests which fails b/c I don't have the proper credentials installed. This could be fixed but isn't straight-forward enough to include in this PR.

PR #235, opened 2019-06-26
Fix medication reports bug

Erin-Mounts:
It turns out no explicit migration is necessary:
- It's been saving all Reports to the /v4 endpoint all along; it's only been pulling from /v3 for date-only reports
- All Reports get refreshed from Bridge (now via the /v4 endpoint) at app startup so after updating to the new BridgeSDK they will automatically get cached into the correct list


BridgeSDK/BridgeAPI/BridgeAPITypes/MachineGeneratedDoNotEdit/_SBBStudyParticipant.m
```diff
@@ -107,8 +107,13 @@ - (instancetype)initWithManagedObject:(NSManagedObject *)managedObject objectMan
     NSString *password = cacheManager.encryptionKey;
     if (password) {
         NSData *plaintext = [RNDecryptor decryptData:managedObject.ciphertext withPassword:password error:nil];
-        NSDictionary *jsonDict = [NSJSONSerialization JSONObjectWithData:plaintext options:0 error:NULL];
-        self = [self initWithDictionaryRepresentation:jsonDict objectManager:objectManager];
+        if (error && !plaintext) {
```
syoung-smallwisdom: ? Um.... super confused by this. Where is `error` defined. It's not being used by the decryptor.

Also, shouldn't this fail if there is an error OR plaintext is nil? Otherwise, it will crash below where you are trying to serialize a `nil` value.

AND should the NSLog be wrapped in `#if DEBUG` so that you aren't submitting a password to the console log?

BridgeSDK/BridgeAPI/BridgeAPITypes/MachineGeneratedDoNotEdit/_SBBUserProfile.m
```diff
@@ -271,8 +271,13 @@ - (instancetype)initWithManagedObject:(NSManagedObject *)managedObject objectMan
...
         NSData *plaintext = [RNDecryptor decryptData:managedObject.ciphertext withPassword:password error:nil];
-        NSDictionary *jsonDict = [NSJSONSerialization JSONObjectWithData:plaintext options:0 error:NULL];
-        self = [self initWithDictionaryRepresentation:jsonDict objectManager:objectManager];
+        if (error && !plaintext) {
+            NSLog(@"Error decrypting %@ with password '%@': %@", self.class.entityName, password, error);
+            self = nil;
+        } else {
+            NSDictionary *jsonDict = [NSJSONSerialization JSONObjectWithData:plaintext options:0 error:NULL];
+            self = [self initWithDictionaryRepresentation:jsonDict objectManager:objectManager];
+        }
```
syoung-smallwisdom: Same question as above

BridgeSDK/mogenerator_ponso_templates/templates/machine.m.motemplate
```diff
@@ -489,8 +489,13 @@ static id dynamicGetterIMP(id self, SEL _cmd)
...
         NSData *plaintext = [RNDecryptor decryptData:managedObject.<$userInfo.encrypted$> withPassword:password error:nil];
-        NSDictionary *jsonDict = [NSJSONSerialization JSONObjectWithData:plaintext options:0 error:NULL];
-        self = [self initWithDictionaryRepresentation:jsonDict objectManager:objectManager];
+        if (error && !plaintext) {
+            NSLog(@"Error decrypting %@ with password '%@': %@", self.class.entityName, password, error);
+            self = nil;
+        } else {
+            NSDictionary *jsonDict = [NSJSONSerialization JSONObjectWithData:plaintext options:0 error:NULL];
+            self = [self initWithDictionaryRepresentation:jsonDict objectManager:objectManager];
+        }
```
syoung-smallwisdom: same

BridgeSDK/BridgeAPI/BridgeAPITypes/MachineGeneratedDoNotEdit/_SBBStudyParticipant.m
Erin-Mounts: ok check again?

PR #240, opened 2019-09-19
Comment out verbose logging about saving to core data

syoung-smallwisdom:


BridgeSDK/Internal/SBBCacheManager.m
```diff
@@ -726,7 +726,7 @@ - (void)saveCacheIOContext
                 NSLog(@"Error saving cache manager's managed object context. If this is a validation error and at some later time the context is successfully saved, you can ignore this error:\n%@",  error);
```
Erin-Mounts: With that message commented out, this one becomes (usually unnecessarily) alarming. Instead please wrap both of them inside #if DEBUG_CACHE_MANAGER ... #endif blocks.

PR #241, opened 2019-10-15
Fix reauth race condition

Erin-Mounts:
Time to bite the bullet and prevent multiple concurrent auth (reauth/signIn) calls to Bridge. Instead, if one is in progress, queue up subsequent ones' completion handlers to be called when the one call completes.

Also, fixing an outstanding issue with falling back to stored credentials w/password that has been causing crashes requiring delete/re-install (which only affected us internally on mPower 2 if we were using test external ids to sign in).


BridgeSDK/BridgeAPI/SBBAuthManager.m
```diff
@@ -548,6 +591,11 @@ - (void)handleSignInWithCredential:(NSString *)credentialKey value:(id<SBBJSONVa
             keysAndValues[self.reauthTokenKey] = reauthToken;
         }
         
+        if (credentialKey && credentialValue) {
+            keysAndValues[self.credentialKeyKey] = credentialKey;
+            keysAndValues[self.credentialValueKey] = credentialValue;
+        }
```
syoung-smallwisdom: Why is this being added? Can you add a comment? Thanks

Also, looks like this is actually set to the keychain using a different queue from the auth queue? Is there a potential for stuff to get clobbered?

BridgeSDK/BridgeAPI/SBBAuthManager.m
```diff
@@ -617,6 +684,12 @@ - (NSURLSessionTask *)signInWithCredential:(NSString *)credentialKey value:(id<S
         return nil;
     }
     
+    // If an auth/reauth call is already in progress, just queue up the completion handler and return the
+    // existing NSURLSessionTask object.
+    if ([self testAndSetAuthCallInProgressWithCompletion:completion]) {
+        return self.inProgressAuthURLSessionTask;
```
syoung-smallwisdom: Does the access to the session task also need to be synchronous?

BridgeSDK/BridgeAPI/SBBAuthManagerInternal.h
```diff
@@ -73,6 +73,15 @@ extern NSString * const kSBBAuthResetAPI;
 @property (nonatomic, strong) id<SBBObjectManagerProtocol> objectManager;
 @property (nonatomic, strong) SBBUserSessionInfo *placeholderSessionInfo;
 
+// Keep track of whether an auth/reauth call to Bridge is currently in progress.
+@property (nonatomic, assign) BOOL authCallInProgress;
```
syoung-smallwisdom: Should these be readonly?

BridgeSDK/BridgeAPI/SBBAuthManagerInternal.h
```diff
@@ -73,6 +73,15 @@ extern NSString * const kSBBAuthResetAPI;
...
+
+// Keep track of the URL session task for an in-progress auth/reauth call to Bridge.
+@property (nonatomic, assign) NSURLSessionTask *inProgressAuthURLSessionTask;
```
syoung-smallwisdom: Shouldn't this be `weak` rather than `assign`? Sorry, can't remember.

Erin-Mounts: Whoops. In fact it should be `strong`

BridgeSDK/BridgeAPI/SBBAuthManagerInternal.h
Erin-Mounts: They're "private" (internal header file not exposed in the module) so I wasn't worried about that... should I be?

BridgeSDK/BridgeAPI/SBBAuthManager.m
Erin-Mounts: No, because access to the flag returned by testAndSetAuthCall...etc. is serialized so only one thread (the one that actually flipped it from "no" to "yes") can ever get past this to eventually change its value.

On the other hand, it turns out I'm never actually setting this value anywhere... oops.

Erin-Mounts: Actually on second thought, I do need to make sure that one thread can't clear it in between the test-and-set and this return. Will fix that. 👍

BridgeSDK/BridgeAPI/SBBAuthManager.m
Erin-Mounts: Adding a comment.

As for the other thing, until the cleanup code executes at the bottom of this method, this is still protected by the "one-auth-call-to-Bridge-at-a-time" serialization so it won't get called on multiple threads.

BridgeSDK/BridgeAPI/SBBAuthManager.m
Erin-Mounts: I ended up just removing the inProgressAuthURLSessionTask property because it's not really needed, and it probably makes sense for only the "original" auth/reauth call that kicked off the Bridge request to have access to the session task anyway.

BridgeSDK/BridgeAPI/SBBAuthManagerInternal.h
Erin-Mounts: went ahead with the suggestion anyway--doesn't hurt and might prevent a hypothetical future obj-c runtime twiddler from getting themselves in trouble if nothing else.

BridgeSDK/BridgeAPI/SBBAuthManagerInternal.h
Erin-Mounts: or maybe just not exist ;)

BridgeSDK/BridgeAPI/SBBAuthManagerInternal.h
syoung-smallwisdom: Yeah, when someone else eventually inherits this code, readonly tells that dev that this *should* only be accessible by the auth manager. Or if it is exposed as internal so that tests can muck with it, might add a comment to that effect? Don't know, pretty edge case. ;)

PR #242, opened 2019-10-16
Fix integration tests

Erin-Mounts:


BridgeSDKTests/SBBParticipantManagerIntegrationTests.m
```diff
@@ -284,10 +287,50 @@ - (void)testSetSharingScope
...
+    [self waitForExpectationsWithTimeout:5.0 handler:^(NSError *error) {
+        if (error) {
+            NSLog(@"Timeout creating external identifier %@: %@", externalIdentifier, error);
+            externalIdentifier = nil;
+        }
+    }];
+
+    if (!externalIdentifier) {
+        // If we couldn't create the externalIdentifier, just fail & bail (we've already logged it)
+        XCTAssert(NO);
```
syoung-smallwisdom: Does Obj-c have `XCTFail(@"Impossible code path");` ? It's always nice to be able to look at the travis logs and see a message for why it failed. Travis doesn't tell you *where* it failed.

BridgeSDKTests/SBBParticipantManagerIntegrationTests.m
```diff
@@ -284,10 +287,50 @@ - (void)testSetSharingScope
...
+        return;
+    }
+    
     XCTestExpectation *expectSetIdentifier = [self expectationWithDescription:@"set external identifier"];
-    NSString *externalIdentifier = @"test-external-identifier";
     [SBBComponent(SBBParticipantManager) setExternalIdentifier:externalIdentifier completion:^(id  _Nullable responseObject, NSError * _Nullable error) {
-        XCTAssert(!error, @"Server accepted external identifier");
+        XCTAssert(!error, @"Server rejected external identifier %@", externalIdentifier);
```
syoung-smallwisdom: `XCTAssertNil()` ? Is that macro available?

BridgeSDKTests/SBBParticipantManagerIntegrationTests.m
```diff
@@ -323,6 +367,33 @@ - (void)testSetExternalIdentifier
...
+        NSString *endpoint = [NSString stringWithFormat:@"%@/%@", DEV_EXTERNAL_ID_API, externalIdentifier];
+        [bridgeMan delete:endpoint headers:headers parameters:externalIdObject completion:^(NSURLSessionTask *task, id responseObject, NSError *error) {
+            if (error) {
+                NSLog(@"Failed to delete external ID %@\nError:%@\nResponse:%@", externalIdentifier, error, responseObject);
+                externalIdentifier = nil;
+            } else {
+                NSLog(@"Deleted test external ID %@", externalIdentifier);
+            }
+        }];
+    }];
```
syoung-smallwisdom: Should probably throw an assert if this fails, too. Just so you know that something went wrong.

BridgeSDKTests/SBBParticipantManagerIntegrationTests.m
```diff
@@ -284,10 +287,50 @@ - (void)testSetSharingScope
...
+            NSLog(@"Created test external ID %@", externalIdentifier);
+        }
+        
+        [expectCreatedIdentifier fulfill];
+    }];
+    
+    [self waitForExpectationsWithTimeout:5.0 handler:^(NSError *error) {
+        if (error) {
+            NSLog(@"Timeout creating external identifier %@: %@", externalIdentifier, error);
+            externalIdentifier = nil;
```
syoung-smallwisdom: I think how expectations work has changed and you need to include an assert here if there was an error.

BridgeSDKTests/SBBParticipantManagerIntegrationTests.m
```diff
@@ -284,10 +287,50 @@ - (void)testSetSharingScope
...
+      @"substudyId": substudyId,
+      @"type": @"ExternalIdentifier"
+      };
+    
+    XCTestExpectation *expectCreatedIdentifier = [self expectationWithDescription:@"created external identifier"];
+    id<SBBBridgeNetworkManagerProtocol> bridgeMan = SBBComponent(SBBBridgeNetworkManager);
+    [bridgeMan post:DEV_EXTERNAL_ID_API headers:headers parameters:externalIdObject completion:^(NSURLSessionTask *task, id responseObject, NSError *error) {
+        if (error) {
+            NSLog(@"Failed to create external ID %@\nError:%@\nResponse:%@", externalIdentifier, error, responseObject);
+            externalIdentifier = nil;
```
syoung-smallwisdom: Why log rather than assert?

Erin-Mounts: good point

BridgeSDKTests/SBBParticipantManagerIntegrationTests.m
Erin-Mounts: 👍

Erin-Mounts: ...actually remembered the reason I didn't do this: I want to make sure the test run doesn't halt prematurely so as much as possible of the test accounts etc. that it sets up in the test study can get cleaned up after each run even when there are failures. I still go in to the test study periodically to clean up stuff that got left behind when it hits an assert in the Bridge or iOS SDK, but it gets out of hand quickly if I don't let these continue after errors.

BridgeSDKTests/SBBParticipantManagerIntegrationTests.m
Erin-Mounts: see other comment about log vs assert

PR #250, opened 2019-11-09
Fix orphan file explosion

Erin-Mounts:
- Optimize the code that goes through finding and uploading orphaned files to be performant even with thousands of such files
- Add a pre-built configuration of OpenSSL that works with code profiling


syoung-smallwisdom:
Looks good

PR #256, opened 2020-11-09
Openssl1 1 1 xcode12

Erin-Mounts:
The important changes are to openssl.h (list of header files to include changed between 1.0.2 and 1.1.1), and the pre-built libopenssl.a binaries. Everything else is pretty much side effects of upgrading openssl. I also made changes to CMSSupport directly and pushed them without review (sorry), which is how the binaries get pre-built, but if you want to have a look at that project the important changes were to the build-libssl.sh script.


syoung-smallwisdom (approved):
Looks good. Can you also bump the minor marketing version? That makes using this with Carthage more apparent. Thanks!

PR #257, opened 2020-11-17
Openssl xcframework

syoung-smallwisdom:
This will build and archive the BridgeAppExample without having to use legacy settings. That said, I was still looking into what kind of code reorg would be needed to support SwiftPM. Looks like it might be a fair amount b/c of the submodules.


Erin-Mounts (approved):
looks good--I suppose we can worry about being able to step into openssl code in the debugger if we ever end up needing that ;)

syoung-smallwisdom:
yes, I couldn't find any literature on building a binary *and* allowing debug of it. ¯\_(ツ)_/¯

PR #258, opened 2020-12-05
Switch to managing dependencies with Swift PM

syoung-smallwisdom:
https://github.com/Sage-Bionetworks/SageResearchCatalog-Apple/pull/1 has examples of how to set up an application to use Carthage and SwiftPM. For the time being, I think we need to continue supporting both.

This PR does a couple of things:
1. Reference the XCFramework (static) built into the CMSSupport/OpenSLL project using the Swift PM binary.
2. Build an XCFramework (dylib) for BridgeSDK
3. Add a Package.swift manifest for the XCframework that includes the OpenSSL dependency.

I choose this structure so that the BridgeSDK framework doesn't have to be tweaked and all the header paths don't have to be changed to support the "Swift Package" way of setting them up. This structure also allows us to set up a version of the apps that will allow us to set into the BridgeSDK framework for debugging (though its a pain to set up).

Following this PR, I will be submitting one for BridgeApp-Apple-SDK that points at the versioned repo (rather than my fork).


Erin-Mounts (approved):
I will take your word for it.

PR #259, opened 2020-12-05
Fix warnings

syoung-smallwisdom:
So... yeah. BridgeApp was failing the automated testing b/c there were too many warnings in BridgeSDK. I went ahead and fixed the ones that were "easy" although this means that I modified the generated files. I figure we can fight that battle another day when/if we need to add to the model. Mostly, I just want to see if I can get BridgeApp to build so MTB can use swiftPM.


Erin-Mounts (approved):
I've put in a pull request on this pull request branch to update the mogenerator templates appropriately. (also updated the sage fork of mogenerator, master branch, to support the template changes, so you'd need to pull down and build the latest from there before running the BridgeObjects target.) Once you merge that into your branch feel free to merge your branch to Sage master.

PR #260, opened 2020-12-07
Fix broken archival process 

syoung-smallwisdom:
This fixes the archival process for SageResearchCatalog. Apparently, I had that right the first time and the broken bit was in travis. *sigh*


Erin-Mounts (approved):
👍

PR #261, opened 2020-12-08
Use --deep code signing flag and submodule for OpenSSL

syoung-smallwisdom:
*sigh* So... yeah. I hope this works. I have tested building the project using travis and the BridgeApp project, as well as uploading the distribution of the SageResearchCatalog project - both using Swift PM and using Carthage. I think that I've finally figured out the settings for deep copying the openSSL *into* the BridgeSDK binary so that it's built as a part of it (and thus the code is signed as a part of it). This set up allows either using Carthage *or* using Swift Package Manager.

This PR will need to be tagged *before* resolving https://github.com/Sage-Bionetworks/BridgeApp-Apple-SDK/pull/196


Erin-Mounts (approved):
👍

